PR #3, opened 2021-11-19
Add sections to navigation

IanVS:
This organizes our pages into three main sections, Overview, Guides, and Configuration Reference

image

The styles here don't matter too much, since they won't be used on defined.net where they'll be published.  

Organizing the pages into sections can be done via the CMS:

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,9 @@ To add a non-English translation of the docs, open the CMS and add a new locale 
 
 ## Architecture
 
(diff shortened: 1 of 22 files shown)


bgolat (approved):

IanVS:
The curly quote issue isn't really a problem with the quotes, it's an issue with this version of astro which is putting our meta tag setting utf-8 in the wrong place.  We can revert 3756ad9 once that's been addressed.

IanVS:
Main is a bit broken right now, going to merge this one and move forward from there.

PR #7, opened 2021-12-02
Add table of contents to right sidebar

IanVS:
This adds a table of contents for the current page to the right side of the page.  It doesn't indicate the active section, partly because that's a bit buggy and hard to do well and I wanted to keep the JS to a minimum.  It only shows the second-level headlines on the page.

image


bgolat:
The heading font for "on this page" isn't displaying properly for me.  Other than that, it looks good.
image

bgolat (approved):

PR #8, opened 2021-12-09
Revert font optimizations to fix italics

vbscott:
Some of the changes to how fonts are handled broke italics for Chrome so this diff is to revert some of those changes.


IanVS (approved):
:+1:

We can give variable fonts another try some other time, once browser support gets a bit better, maybe.

PR #15, opened 2021-12-20
Move pages inside `/nebula` route

IanVS:
This addresses an issue with internal links between docs pages.  We host these nebula docs at `www.defined.net/nebula`, whereas we were serving them in this repo directly out of the root.  This PR unifies the structure, so that we can use root-relative links between the pages, and they will work here when building/previewing the docs in this repo.  

Also, since we currently are not using localized pages on defined.net, this adds the ability to use urls like `/nebula/overview` directly, and it will work in both defined.net as well as in this repo.

### DIFF
--- a/src/components/LeftSidebar/LeftSidebar.astro
+++ b/src/components/LeftSidebar/LeftSidebar.astro
@@ -2,7 +2,7 @@
 import {getSections} from '../../get-page-links'
 import NavLink from '../Nav/NavLink.astro';
 
-const {currentPage} = Astro.props;
+const currentPage = Astro.request.canonicalURL.pathname;
 const allPages = Astro.fetchContent('../../data/**/*.md');
 const sections = getSections({currentPage, allPages})
 
--- a/src/components/MobileNav.astro
+++ b/src/components/MobileNav.astro
@@ -1,7 +1,7 @@
 ---
 import {getPageLinks} from '../get-page-links'
 const allPages = Astro.fetchContent('../data/**/*.md');
-const currentPage = Astro.request.url.pathname.replace(/\/$/, '');
+const currentPage = Astro.request.canonicalURL.pathname;
 const pages = getPageLinks({currentPage, allPages});
 
 const currentPageTitle = pages.find(p => p.href === currentPage)?.title
--- a/src/pages/nebula/[slug].astro
+++ b/src/pages/nebula/[slug].astro
@@ -1,0 +1,49 @@
+---
+import Layout from '../../layouts/MainLayout.astro';
+import ConfigReference from '../../components/ConfigReference.astro';
+import {DEFAULT_LOCALE} from '../../languages';
+
+/**
+ * This file is used for urls without a locale, which means they are for the default language.
+ */
+
+export async function getStaticPaths() {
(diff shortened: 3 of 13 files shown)


src/pages/nebula/[slug].astro
```diff
@@ -0,0 +1,49 @@
...
+}
+
+const { 
+  /** Contains the frontmatter from the page data, including title, summary, etc. */
+  page, 
+} = Astro.props
+
+const isConfigRefPage = page.slug === 'config';
+
+if (isConfigRefPage) {
```
Comment (briankelly) [inline src/pages/nebula/[slug].astro:38]: I'm not sure what this if block is doing, but I noticed that when running the `nebula-docs` server locally, that I get a 500 when viewing the config page linked from the sidebar (`/nebula/en/config`).

It does load if I attempt to view `/nebula/config`, but all the other pages are okay when `en` is included in the URI.

I tried commenting out this if block and that prevented the 500 from showing, but I'm not sure what the side-effects might be.

briankelly (changes requested):
fix config page so that it loads from local server

- `/nebula/en/config` throws a 500
- `/nebula/config` works, but it's not linked in the sidebar (and the other guides include `/en`)

src/pages/nebula/[slug].astro
Comment (IanVS) [reply src/pages/nebula/[slug].astro:38]: Thanks for the catch, I pushed up a fix.

IanVS:
Alrighty, I've pushed a change to handle https://github.com/withastro/astro/issues/1415.  I think this should be ready for a final lookover.

src/components/MobileNav.astro
```diff
@@ -1,7 +1,7 @@
 ---
 import {getPageLinks} from '../get-page-links'
 const allPages = Astro.fetchContent('../data/**/*.md');
```
briankelly, line 3: I noticed that the **Guides** index is not included in the mobile drop-down navigation.

Is this expected?

briankelly: image

src/components/LeftSidebar/LeftSidebar.astro
```diff
@@ -2,7 +2,7 @@
 import {getSections} from '../../get-page-links'
 import NavLink from '../Nav/NavLink.astro';
 
-const {currentPage} = Astro.props;
+const currentPage = Astro.request.canonicalURL.pathname;
```
briankelly, line 5: When in Desktop view, **Overview**, **Guides**, and **Configuration Reference** get highlighted correctly in the left-sidebar, but _Quick Start_ and _ Extend network access beyond overlay hosts_ do not. 

Is this expected?

Note: Those pages _do_ get highlighted correctly when viewing as part of the www site.

Expected: 
image

Not highlighted:
image

briankelly (commented):
All pages load, but I have a few questions about nav highlights.

briankelly (approved):

src/components/LeftSidebar/LeftSidebar.astro
briankelly: expected behavior - resolving

src/components/MobileNav.astro
briankelly: expected behavior - resolving

PR #23, opened 2022-04-14
Upgrade astro to `0.25` and integrate migration changes

jasikpark:
This upgrades `astro` to `0.25` and makes some changes referenced in the migration guide: https://docs.astro.build/en/migrate/#migrate-to-v025

We're moving from `0.21` to `0.25` here


IanVS:
It looks like the blocking issue has been fixed and released in https://github.com/withastro/compiler/releases/tag/%40astrojs%2Fcompiler%400.14.2. Do we need to wait for a new release of astro, to pull this in?  Or, can we upgrade the compiler independently?

PR #27, opened 2022-05-27
Update nebula config reference

jasikpark:
Related to https://github.com/DefinedNet/nebula-docs/issues/19

You can see the deploy preview in `www`

It'd be helpful to get some confirmation on which things I've labeled as reloadable and not reloadable

### DIFF
--- a/src/components/ConfigReference.astro
+++ b/src/components/ConfigReference.astro
@@ -20,6 +20,8 @@ const configOptions = options[locale] ? options[locale].options : fallbackOption
     <div>
       <ConfigOption
         isRequired={option.required}
+        isReloadable={option.reloadable}
+        defaultValue={option.default}
         description={option.description ?? fallbackOption.description}
         example={option.example ?? fallbackOption.example}
       >
@@ -27,7 +29,7 @@ const configOptions = options[locale] ? options[locale].options : fallbackOption
           {option.name}
           <a
             aria-hidden="true"
-            tabindex="-1"
+            tabindex={-1}
             class="heading-link no-underline"
             href={`#${option.name}`}
           >
@@ -42,6 +44,8 @@ const configOptions = options[locale] ? options[locale].options : fallbackOption
           return (
             <ConfigOption
               isRequired={suboption.required}
+              isReloadable={option.reloadable}
+              defaultValue={option.default}
               description={suboption.description ?? fallbackSuboption.description}
               example={suboption.example ?? fallbackSuboption.example}
             >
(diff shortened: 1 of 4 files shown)


src/data/config-reference.md
```diff
@@ -204,8 +204,8 @@ en:
         - name: port
           default: 0
           description: >
-            port is the UDP port nebula should use on a host. setting this to 0
-            will dynamically assign a port number.
+            port is the UDP port nebula should use on a host. setting this to `0`
```
IanVS, line 207: If the default is `0`, does "setting this to `0`" really make a difference?  Or, should we say that leaving it at the default dynamically....

JohnMaguire: This is the existing text that I believe comes originally from the `config.yml`. "setting this to `0`" makes sense to me, and makes it clear what happens if you explicitly set it, e.g. through an ansible playbook

```
port: {{ port | default(0) }}
```

IanVS: I would prefer language like `A value of "0" (the default) will dynamically assign a port number.`

As it is, it makes it sound like the default value is not to dynamically assign a port number, and that setting the value will cause that to happen.  But, I'm 1/5, feel free to ignore.

IanVS:
You'll need to change the CMS config as well, to expose these settings.

jasikpark:
> You'll need to change the CMS config as well, to expose these settings.

Mind pairing w/ me on this? I'm not sure how or where I'd pull them in

IanVS (changes requested):
I think we might want to find a way to explain what `reloadable` means, and we should definitely get a review on the content from @nbrownus.  But I only had one blocker here.

src/components/ConfigReference.astro
```diff
@@ -35,6 +37,8 @@ const configOptions = options[locale] ? options[locale].options : fallbackOption
         
         return (<ConfigOption
           isRequired={suboption.required}
+          isReloadable={option.reloadable}
+          defaultValue={option.default}
```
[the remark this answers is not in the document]
jasikpark: oops, I was copying over from the `www` PR

jasikpark:
Yeah, I don't fully understand what sending `SIGHUP` would mean exactly... I guess it's as much as:

> While nebula is running for `RELOADABLE` properties you can edit the values in the config and signal the running nebula process with `SIGHUP` to have nebula update those config properties.
> ex:
> run `ps aux | grep nebula` to find nebula's PID & then run `kill -s HUP $PID`

IanVS (approved):

jasikpark:
@nbrownus do you think this could be merged if we yoink out the `required` pills for the moment and only add the `default` pills + the updated docs?

nbrownus:
I will need to re-review the copy but in general yes, I think we need to workshop the reloadable indicators more.

jasikpark:
> I will need to re-review the copy but in general yes, I think we need to workshop the reloadable indicators more.

Ok, I'll try to push up a commit that removes the "reloadable" indicators before too long

jasikpark:
Replaced by #42

PR #32, opened 2022-07-27
Add a hint on how to generate SSH host key

JohnMaguire:
Screen Shot 2022-07-27 at 12 44 55 PM

### DIFF
--- a/src/data/config-reference.md
+++ b/src/data/config-reference.md
@@ -243,15 +243,20 @@ en:
         nebula node.
       example: 'local_range: "172.16.0.0/24"'
     - name: sshd
-      description: sshd enables nebula's built-in debugging console, which can be
+      description: >-
+        sshd enables nebula's built-in debugging console, which can be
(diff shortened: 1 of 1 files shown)


IanVS (approved):
LGTM, although @bgolat, maybe we should change the styles a bit so the inline code blocks don't overlap like this.

JohnMaguire:
> LGTM, although @bgolat, maybe we should change the styles a bit so the inline code blocks don't overlap like this.

The spacing before the inline code block looks a little off to me too. (More space inside the code block than outside.)

bgolat:
> LGTM, although @bgolat, maybe we should change the styles a bit so the inline code blocks don't overlap like this.

That screenshot is using old code styles for some reason.  We don't use those fonts/colors/spacing anymore.

IanVS:
Here's how it looks when built into the website: 

image

We just don't have all the styles fully synced into this project when built outside of the website.

PR #39, opened 2022-08-18
Replace astro with docusaurus

IanVS:
This will simplify the creation and publishing of nebula docs, using a popular open-source documentation tool, docusaurus, rather than custom built astro components and styles.


jasikpark:
gotta update the netlify toml to change the baseurl for previews

docusaurus.config.cjs
```diff
@@ -0,0 +1,136 @@
...
+                label: 'Home',
+                href: 'https://defined.net',
+              },
+              {
+                label: 'Blog',
+                href: 'https://defined.net/blog',
+              },
+              {
+                label: 'Docs',
+                href: 'https://docs.defined.net/dn',
```
jasikpark, line 118: This won't be valid until we deploy DN docs, but it seems fine for now

postcss.config.js
```diff
@@ -0,0 +1,6 @@
+const postcssNesting = require('postcss-nesting');
+const autoprefixer = require('autoprefixer');
+
+module.exports = {
+  plugins: [autoprefixer(), postcssNesting()],
```
jasikpark, line 5: where did you find this to add? should this not already be added inside docusaurus?

public/make-scrollable-code-focusable.js
```diff
@@ -1,3 +0,0 @@
-Array.from(document.getElementsByTagName("pre")).forEach((element) => {
```
jasikpark, line 1: I appreciate that this is dealt w/ by docusaurus by default, rather than requiring a custom script

src/components/Redirect.astro
```diff
@@ -1,5 +0,0 @@
----
-const { target } = Astro.props;
----
-
-<head><meta http-equiv="refresh" content={`0; url=${target}`} /></head>
```
jasikpark, line 5: where did we use this instead of netlify redirects before?

src/css/base.css
```diff
@@ -0,0 +1,61 @@
...
+  --dn-color-gray-50: hsl(var(--dn-color-gray-hs), 50%);
+  --dn-color-gray-40: hsl(var(--dn-color-gray-hs), 40%);
+  --dn-color-gray-30: hsl(var(--dn-color-gray-hs), 30%);
+  --dn-color-gray-20: hsl(var(--dn-color-gray-hs), 20%);
+  --dn-color-gray-15: hsl(var(--dn-color-gray-hs), 15%);
+  --dn-color-gray-10: hsl(var(--dn-color-gray-hs), 10%);
+  --dn-color-gray-05: hsl(var(--dn-color-gray-hs), 5%);
+  /* darkest */
+
+  --dn-color-green-hs: 135, 64%;
```
jasikpark, line 47: why add these base colors w/o the full spectrum of lightnesses?

jasikpark:
Screen Shot 2022-08-18 at 17 58 14

"back to main menu" seems broken when we don't have a full website and are hosting docs at "/" from the perspective of docusaurus

jasikpark:
image

a cool thing is that we can eventually render a light mode version of this and use github style light/dark mode selection for it?

or just using `<ThemedImage alt={} sources={}/>` https://docusaurus.io/docs/markdown-features/assets#themed-images

src/css/base.css
IanVS: Because they're only used in the prism themes, and it's easy enough to add the lightness needed there.

src/components/Redirect.astro
IanVS: It was part of the i18n setup.

src/css/base.css
jasikpark: eh

src/components/Redirect.astro
jasikpark: right, fair

postcss.config.js
IanVS: Auto prefixed is used by default, but not nesting.

jasikpark (approved):

Issue #41, opened 2022-08-22
Improve config documentation for logging, stats, preferred_ranges, routines


jasikpark:
including:

- logging
	  - format
	  - level
	  - timestamp_format
- stats
	- type
	- protocol
	- host
	- prefix
	- namespace
	- subsystem
	- path
	- interval
	- lighthouse_metrics
	- message_metrics
- preferred_ranges
- routines

I'm sure that some of this can be pulled from https://github.com/slackhq/nebula/blob/master/examples/config.yml

jasikpark:
Fixed by #42

PR #42, opened 2022-08-22
Improve Config Reference docs

jasikpark:
Improves #41 
Replaces #27 

This PR lists default / updates docs for:

- pki
- lighthouse
- listen
- punchy
- sshd
- tun
- logging
- firewall
- preferred_ranges
- routines
- stats

config groups in one way or another

https://deploy-preview-42--nebula-docs-dn.netlify.app/nebula/config

### DIFF
--- a/docs/config/cipher.mdx
+++ b/docs/config/cipher.mdx
@@ -6,8 +6,7 @@ sidebar_position: 6
 
 :::danger
 
-This value must be identical on ALL NODES/LIGHTHOUSES. We do not/will not support use of different ciphers
-simultaneously!
+This value must be identical on ALL NODES/LIGHTHOUSES. Nebula does not support use of different ciphers simultaneously!
 
 :::
 
--- a/docs/config/cipher.mdx
+++ b/docs/config/cipher.mdx
@@ -4,8 +4,12 @@ sidebar_position: 6
 
 # cipher
 
-**IMPORTANT: this value must be identical on ALL NODES/LIGHTHOUSES. We do not/will not support use of different ciphers
-simultaneously!**
+:::danger
+
+This value must be identical on ALL NODES/LIGHTHOUSES. We do not/will not support use of different ciphers
+simultaneously!
+
+:::
 
 `cipher` allows you to choose between the available ciphers for your network. The default is `aes`, and the available
 options are `chachapoly` or `aes`.
--- a/docs/config/firewall.mdx
+++ b/docs/config/firewall.mdx
@@ -68,7 +68,6 @@ firewall:
     tcp_timeout: 12m
     udp_timeout: 3m
     default_timeout: 10m
-    max_connections: 100000
 
   outbound:
     # Allow all outbound traffic from this node
--- a/docs/config/firewall.mdx
+++ b/docs/config/firewall.mdx
@@ -61,23 +61,26 @@ The possible fields of a firewall rule are:
 Logical evaluation is roughly: port AND proto AND (ca_sha OR ca_name) AND (host OR group OR groups OR cidr).
 
 ```yml
-
 # Nebula security group configuration
 
 firewall:
-conntrack:
-tcp_timeout: 12m
-udp_timeout: 3m
-default_timeout: 10m
-max_connections: 100000
+  conntrack:
+    tcp_timeout: 12m
+    udp_timeout: 3m
+    default_timeout: 10m
+    max_connections: 100000
 
-outbound: # Allow all outbound traffic from this node - port: any
-proto: any
-host: any
+  outbound:
+    # Allow all outbound traffic from this node
+    - port: any
+      proto: any
+      host: any
 
-inbound: # Allow icmp between any nebula hosts - port: any
-proto: icmp
-host: any
+  inbound:
+    # Allow icmp between any nebula hosts
+    - port: any
+      proto: icmp
+      host: any
 
     # Allow tcp/443 from any host with BOTH laptop and home group
     - port: 443
@@ -104,3 +107,14 @@ outbound:
 
 At a minimum, it is recommended to enable ICMP so that `ping` can be used to verify connectivity. Additionally, if
 enabling the built-in Nebula SSH server, you may wish to grant access over the Nebula network via firewall rules.
+
+## firewall.conntrack
+
+Settings for the Connection Tracker.
+
+```yml
+conntrack:
+  tcp_timeout: 12m
+  udp_timeout: 3m
+  default_timeout: 10m
+```
--- a/docs/config/lighthouse.mdx
+++ b/docs/config/lighthouse.mdx
@@ -34,15 +34,15 @@ lighthouse:
 
 <Pill className="mb-24">Default: False</Pill>
 
-am_lighthouse is used to enable lighthouse functionality for a node. This should ONLY be `true` on nodes you have
+`am_lighthouse` is used to enable lighthouse functionality for a node. This should ONLY be `true` on nodes you have
 configured to be lighthouses in your network
 
 ## lighthouse.serve_dns
 
 <Pill className="mb-24">Default: False</Pill>
 
-serve_dns optionally starts a DNS listener that responds to `A` and `TXT` queries and can even be delegated to for name
-resolution by external DNS hosts.
+`serve_dns` optionally starts a DNS listener that responds to `A` and `TXT` queries and can even be delegated to for
+name resolution by external DNS hosts.
 
 The DNS listener can only respond to requests about hosts it's aware of. For this reason, it can only be enabled on
 Lighthouses.
@@ -75,13 +75,13 @@ the [firewall](/config/firewall)._
 
 ## lighthouse.dns
 
-dns is used to configure the address (`host`) and port (`port`) the DNS server should listen on. By listening on the
+`dns` is used to configure the address (`host`) and port (`port`) the DNS server should listen on. By listening on the
 host's Nebula IP, you can make the DNS server accessible only on the Nebula network. Alternatively, listening on
 `0.0.0.0` will allow anyone that can reach the host to make queries.
 
 The default value for `lighthouse.dns.port` is `53` but you must set an ip address.
 
-See the [serve_dns](#lighthouseserve_dns) docs for more information, or the [example Lighthouse config](#) for syntax.
+See the [`serve_dns`](#lighthouseserve_dns) docs for more information, or the [example Lighthouse config](#) for syntax.
 
 ```yml
 dns:
@@ -102,7 +102,7 @@ a large nebula network.
 
 **IMPORTANT: THIS SHOULD BE EMPTY ON LIGHTHOUSE NODES**
 
-hosts is a list of lighthouse hosts this node should report to and query from. The lighthouses listed here should be
+`hosts` is a list of lighthouse hosts this node should report to and query from. The lighthouses listed here should be
 referenced by their **nebula IP**, not by the IPs of their physical network interfaces.
 
 ```yml
@@ -112,7 +112,7 @@ hosts:
 
 ## lighthouse.remote_allow_list
 
-remote_allow_list allows you to control ip ranges that this node will consider when handshaking to another node. By
+`remote_allow_list` allows you to control ip ranges that this node will consider when handshaking to another node. By
 default, any remote IPs are allowed. You can provide CIDRs here with `true` to allow and `false` to deny. The most
 specific CIDR rule applies to each remote. If all rules are "allow", the default will be "deny", and vice-versa. If both
 "allow" and "deny" rules are present, then you MUST set a rule for "0.0.0.0/0" as the default.
@@ -129,10 +129,10 @@ remote_allow_list:
 
 ## lighthouse.local_allow_list
 
-local_allow_list allows you to filter which local IP addresses we advertise to the lighthouses. This uses the same logic
-as `remote_allow_list`, but additionally, you can specify an `interfaces` map of regular expressions to match against
-interface names. The regexp must match the entire name. All interface rules must be either true or false (and the
-default will be the inverse). CIDR rules are matched after interface name rules. Default is all local IP addresses.
+`local_allow_list` allows you to filter which local IP addresses we advertise to the lighthouses. This uses the same
+logic as `remote_allow_list`, but additionally, you can specify an `interfaces` map of regular expressions to match
+against interface names. The regexp must match the entire name. All interface rules must be either true or false (and
+the default will be the inverse). CIDR rules are matched after interface name rules. Default is all local IP addresses.
 
 ```yml
 local_allow_list:
--- a/docs/config/lighthouse.mdx
+++ b/docs/config/lighthouse.mdx
@@ -70,8 +70,12 @@ $ dig +short @192.168.100.1 192.168.100.5 TXT
 "Name: web01" "Ips: [192.168.100.5/24]" "Subnets []" "Groups [servers web]" "NotBefore 2021-06-15 14:19:22 +0000 UTC" "NotAFter 2022-04-28 21:49:15 +0000 UTC" "PublicKey dde33784fb2bbada73f8bf4cafbf9271dc864b770b1e44002f81563856711f7c" "IsCA false" "Issuer 91f795c52f601d3110ee5232b22c13a89a76d3e3fb89bed3c21929c873cb6ec9"
 ```
 
-_NOTE: To allow hosts to make queries against the DNS server over the Nebula network, don't forget to allow access in
-the [firewall](/config/firewall)._
+:::note
+
+To allow hosts to make queries against the DNS server over the Nebula network, don't forget to allow access in the
+[firewall](/config/firewall).
+
+:::
 
 ## lighthouse.dns
 
--- a/docs/config/lighthouse.mdx
+++ b/docs/config/lighthouse.mdx
@@ -85,7 +85,7 @@ host's Nebula IP, you can make the DNS server accessible only on the Nebula netw
 
 The default value for `dns.port` is `53` but you must set an IP address.
 
-See the [`serve_dns`](#lighthouseserve_dns) docs for more information, or the [example Lighthouse config](#) for syntax.
+See the [`serve_dns`](#lighthouseserve_dns) docs for more information.
 
 ```yml
 dns:
--- a/docs/config/lighthouse.mdx
+++ b/docs/config/lighthouse.mdx
@@ -88,10 +88,11 @@ The default value for `dns.port` is `53` but you must set an IP address.
 See the [`serve_dns`](#lighthouseserve_dns) docs for more information.
 
 ```yml
-dns:
-  # The DNS host defines the IP to bind the dns listener to. This also allows binding to the nebula node IP.
-  host: 0.0.0.0
-  port: 53
+lighthouse:
+  dns:
+    # The DNS host defines the IP to bind the dns listener to. This also allows binding to the nebula node IP.
+    host: 0.0.0.0
+    port: 53
 ```
 
 ## lighthouse.interval
@@ -99,14 +100,14 @@ dns:
 <Pill className="mb-24">Default: 10</Pill>
 
 `interval` specifies how often a nebula host should report itself to a lighthouse. By default, hosts report themselves
-to lighthouses once every 60 seconds. Use caution when changing this interval, as it may affect host discovery times in
+to lighthouses once every 10 seconds. Use caution when changing this interval, as it may affect host discovery times in
 a large nebula network.
 
 ## lighthouse.hosts
 
 :::caution
 
-THIS SHOULD BE EMPTY ON LIGHTHOUSE NODES
+This should be empty on lighthouse nodes
 
 :::
 
@@ -114,8 +115,9 @@ THIS SHOULD BE EMPTY ON LIGHTHOUSE NODES
 referenced by their **nebula IP**, not by the IPs of their physical network interfaces.
 
 ```yml
-hosts:
-  - '192.168.100.1'
+lighthouse:
+  hosts:
+    - '192.168.100.1'
 ```
 
 ## lighthouse.remote_allow_list
@@ -126,13 +128,14 @@ specific CIDR rule applies to each remote. If all rules are "allow", the default
 "allow" and "deny" rules are present, then you MUST set a rule for "0.0.0.0/0" as the default.
 
 ```yml
-remote_allow_list:
-  # Example to block IPs from this subnet from being used for remote IPs.
-  '172.16.0.0/12': false
-  # A more complicated example, allow public IPs but only private IPs from a specific subnet
-  '0.0.0.0/0': true
-  '10.0.0.0/8': false
-  '10.42.42.0/24': true
+lighthouse:
+  remote_allow_list:
+    # Example to block IPs from this subnet from being used for remote IPs.
+    '172.16.0.0/12': false
+    # A more complicated example, allow public IPs but only private IPs from a specific subnet
+    '0.0.0.0/0': true
+    '10.0.0.0/8': false
+    '10.42.42.0/24': true
 ```
 
 ## lighthouse.local_allow_list
@@ -143,11 +146,12 @@ against interface names. The regexp must match the entire name. All interface ru
 the default will be the inverse). CIDR rules are matched after interface name rules. Default is all local IP addresses.
 
 ```yml
-local_allow_list:
-  # Example to block tun0 and all docker interfaces.
-  interfaces:
-    tun0: false
-    'docker.*': false
-    # Example to only advertise this subnet to the lighthouse.
-    '10.0.0.0/8': true
+lighthouse:
+  local_allow_list:
+    # Example to block tun0 and all docker interfaces.
+    interfaces:
+      tun0: false
+      'docker.*': false
+      # Example to only advertise this subnet to the lighthouse.
+      '10.0.0.0/8': true
 ```
--- a/docs/config/lighthouse.mdx
+++ b/docs/config/lighthouse.mdx
@@ -2,6 +2,8 @@
 sidebar_position: 3
 ---
 
+import { Pill } from '@components/Pill/Pill';
+
 # lighthouse
 
 ```yml
@@ -30,10 +32,14 @@ lighthouse:
 
 ## lighthouse.am_lighthouse
 
+<Pill className="mb-24">Default: False</Pill>
+
 am_lighthouse is used to enable lighthouse functionality for a node. This should ONLY be `true` on nodes you have
 configured to be lighthouses in your network
 
 ## lighthouse.serve_dns
+
+<Pill className="mb-24">Default: False</Pill>
 
 serve_dns optionally starts a DNS listener that responds to `A` and `TXT` queries and can even be delegated to for name
 resolution by external DNS hosts.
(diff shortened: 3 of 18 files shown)


docs/config/firewall.mdx
```diff
@@ -52,23 +52,26 @@ The possible fields of a firewall rule are:
...
-host: any
-
-inbound: # Allow icmp between any nebula hosts - port: any
-proto: icmp
-host: any
+  conntrack:
+    tcp_timeout: 12m
+    udp_timeout: 3m
+    default_timeout: 10m
+    max_connections: 100000
```
nbrownus, line 62: We can remove `max_connections` as it's not actually used

jasikpark: Can I put up a PR to nebula to rip out the config code as well then? Or do you mean that too

docs/config/lighthouse.mdx
```diff
@@ -30,10 +32,14 @@ lighthouse:
...
 
+<Pill className="mb-24">Default: False</Pill>
+
 am_lighthouse is used to enable lighthouse functionality for a node. This should ONLY be `true` on nodes you have configured to be lighthouses in your network
 
 ## lighthouse.serve_dns
 
+<Pill className="mb-24">Default: False</Pill>
+
 serve_dns optionally starts a DNS listener that responds to `A` and
```
nbrownus, line 43: Depending on the above maybe this is `serve_dns`

docs/config/lighthouse.mdx
```diff
@@ -30,10 +32,14 @@ lighthouse:
 
 ## lighthouse.am_lighthouse
 
+<Pill className="mb-24">Default: False</Pill>
+
 am_lighthouse is used to enable lighthouse functionality for a node. This should ONLY be `true` on nodes you have configured to be lighthouses in your network
```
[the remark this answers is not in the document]
jasikpark: Mind sharing a screenshot? I'm unsure exactly what you mean here

docs/config/firewall.mdx
jasikpark: Removed

docs/config/lighthouse.mdx
jasikpark: Are you recommending cleaning up how we refer to config options? ohhh yeah fair. I'll push up a commit that wraps all of them in backticks / `<code>`

jasikpark: will do

docs/config/listen.mdx
```diff
@@ -2,9 +2,11 @@
 sidebar_position: 4
 ---
 
+import { Pill } from '@components/Pill/Pill';
+
 # listen
 
-listen sets the UDP port Nebula will use for sending/receiving traffic and for handshakes. The default here is 4242. For a lighthouse node, the port should be defined, however using port 0 will dynamically assign a port and is recommended for roaming nodes.
+`listen` sets the UDP port Nebula will use for sending/receiving traffic and for handshakes. For a lighthouse node, the port should be defined, conventionally to `4242`, however using port 0 will dynamically assign a port and is recommended for roaming nodes.
```
nbrownus, line 9: I think we should pull the `listen.port` specific information down to the `listen.port` area, maybe we can simplify the first sentence as well?

> `listen` sets the UDP port Nebula will use for sending and receiving all traffic.

docs/config/lighthouse.mdx
jasikpark: done

docs/config/listen.mdx
```diff
@@ -17,12 +19,20 @@ listen:
...
 ## listen.host
 
+<Pill className="mb-24">Default: 0.0.0.0</Pill>
+
 `host` is the ip of the interface to use when binding the listener. The default is 0.0.0.0, which is what most people should use.
 
 ## listen.port
 
-`port` is the UDP port nebula should use on a host. Setting this to 0 will dynamically assign a port number.
+`port` is the UDP port nebula should use on a host. Setting this to `0` will dynamically assign a port number.
```
nbrownus, line 28: Then here we can say something like

> `port` is the UDP port nebula should use on a host. Setting this to `0` will dynamically assign a port number. Using `0` on lighthouses and relay hosts will likely lead to connectivity issues.

docs/config/listen.mdx
jasikpark: Done

docs/config/listen.mdx
```diff
@@ -17,12 +19,20 @@ listen:
...
+
+## batch
+
+<Pill className="mb-24">Default: 64</Pill>
```
nbrownus, line 32: There are a handful of OS specific settings, maybe a future improvement is to make a pill for it?

docs/config/listen.mdx
jasikpark: Good point w/ the relays/lighthouses

docs/config/listen.mdx
jasikpark: What would you put in such a thing? It's not entirely clear what all of the config values that exist are from the example config file in slackhq/nebula

docs/config/logging.mdx
```diff
@@ -11,3 +13,34 @@ logging:
...
+## logging.disable_timestamp
+
+<Pill className="mb-24">Default: False</Pill>
+
+Disable timestamp logging. Useful when output is redirected to logging system that already adds timestamps.
+
+## logging.timestamp_format
+
+```yml
+# timestamp format is specified in Go time format, see:
```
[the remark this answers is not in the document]
jasikpark: sounds good to me 👍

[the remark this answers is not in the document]
jasikpark: Pushed a diff that implements this in some fashion 👍

docs/config/preferred-ranges.mdx
```diff
@@ -0,0 +1,14 @@
+---
+sidebar_position: 12
+---
+
+# preferred_ranges
+
+Preferred ranges is used to define a hint about the local network ranges, which speeds up discovering the fastest
```
nbrownus: Maybe an improvement:

> Preferred ranges sets the priority order for underlay ip addresses. Two hosts on the same LAN would likely benefit from having their tunnels use the LAN ip addresses rather than the public ip addresses the lighthouse may have learned for them. 

Specifying the default order would also be nice: ipv6 > public ipv4 > private ipv4

docs/config/punchy.mdx
```diff
@@ -15,12 +17,18 @@ punchy:
...
+<Pill className="mb-24">Default: False</Pill>
+
 `respond` means that a node unable to receive handshakes will attempt to initiate a handshake to the host attempting to establish a tunnel, which can be the case when hole punching fails in one direction. This can be extremely useful if one node is behind a difficult nat, such as a symmetric NAT.
 
 ## punchy.delay
 
-`delay` slows down punch responses, which can be helpful for misbehaving NATs or conditions where a NAT router’s conntrack map is unable to handle a race, default is 1 second, respond must be true to take effect.
+<Pill className="mb-24">Default: 1s</Pill>
+
+`delay` slows down punch responses, which can be helpful for misbehaving NATs or conditions where a NAT router’s conntrack map is unable to handle a race, default is 1 second, respond must be true to take effect. Changing this value while running will not affect any in progress operations, only new ones.
```
nbrownus, line 34: I think we can break apart that first sentence a bit

> `delay` slows down punch responses which can be helpful for conditions where a NAT is unable to handle certain race conditions. `respond` must be `true` to take effect.

docs/config/routines.mdx
```diff
@@ -0,0 +1,11 @@
+---
+sidebar_position: 13
+---
+
+# routines
+
```
nbrownus, line 6: We should give some guidance and reword this a bit. 1/2 the number of cpu cores is usually the maximum you would want to set this to. There are many reasons to not set it to the maximum as well.

nbrownus: Also the default pill

jasikpark: Added a tiny bit of guidance on that -- I agree it needs a bit more work though; mostly just grabbed the info from the example config originally

docs/config/tun.mdx
```diff
@@ -39,18 +45,36 @@ Toggles forwarding of multicast packets
...
+    route: 10.0.0.0/16
+```
+
 ## tun.unsafe_routes
 
 **IMPORTANT NOTE: The nebula certificate of the “via” node MUST have the “route” defined as a subnet in its certificate**
 
 Unsafe routes allows you to route traffic over nebula to non-nebula nodes. Unsafe routes should be avoided unless you have hosts/services that cannot run nebula.
+
+```yml
```
nbrownus, line 74: I wonder if the example yaml would benefit from having the full config block to make it more apparent where to copy/paste these things

```yml
tun:
  unsafe_routes:
    - route: 172.16.1.0/24
      via: 192.168.100.99
      mtu: 1300
      metric: 100

File: docs/config/preferred-ranges.mdx (hunk 8 above)
Comment (jasikpark) [reply docs/config/preferred-ranges.mdx]: Does this setting allow specifying `ipv6`, `public ipv4`, `private ipv4` as categories? 

Maybe we should provide examples for how to order those correctly..

Comment (nbrownus) [reply docs/config/preferred-ranges.mdx]: There are no magic words in the config here.

- `::/0` is all ipv6
- `0.0.0.0/0` is all ipv4
- `192.168.0.0/16` would be a specific private ipv4 network

Examples like this would likely be helpful

File: docs/config/tun.mdx (hunk 11 above)
Comment (jasikpark) [reply docs/config/tun.mdx:75]: Might as well 🤷

**Comment by jasikpark** [comment]
Moved `local_range` to the bottom of the list, since it's listed as deprecated

**Comment by jasikpark** [comment]
Currently trying to improve examples for `preferred_ranges`

File: docs/config/tun.mdx (hunk 11 above)
Comment (jasikpark) [reply docs/config/tun.mdx:75]: good point 😅 I don't know at that level...

File: docs/config/preferred-ranges.mdx [hunk 12]
```diff
@@ -8,9 +8,21 @@ sidebar_position: 7
...
 and has been deprecated for `preferred_ranges`
 
 ```yml
 preferred_ranges: ['172.16.0.0/24']
 ```
+
+To set up a host that prefers local network `192.168.0.0/16` first, then `ipv6` then `public ipv4` then `private ipv4`, you would do:
+
+```yml
+preferred_ranges: ['192.168.0.0/16', '::/0', '0.0.0.0/0', 'what do i do here']
```
jasikpark, line 31: I suppose 

https://github.dev/slackhq/nebula/blob/c2259f14a7d230464613c46e36cbe82116fa2de8/remote_list.go#L543-L564

is relevant here? I guess the real idea is that you only specify the few LANs that you prefer rather than customizing the whole cascade, right? + looking at the code the order of specified ranges doesn't matter, it just looks at set inclusion?

nbrownus: https://github.com/slackhq/nebula/blob/master/remote_list.go#L452-L514

jasikpark: Thanks for the link, I've slightly improved the wording / example description... could probably still use work, including describing the sorting algorithm

docs/config/lighthouse.mdx
```diff
@@ -60,28 +66,68 @@ $ dig +short @192.168.100.1 192.168.100.5 TXT
...
+
+:::
 
 ## lighthouse.dns
 
-dns is used to configure the address (`host`) and port (`port`) the DNS server should listen on. By listening on the host's Nebula IP, you can make the DNS server accessible only on the Nebula network. Alternatively, listening on `0.0.0.0` will allow anyone that can reach the host to make queries.
+`dns` is used to configure the address (`host`) and port (`port`) the DNS server should listen on. By listening on the host's Nebula IP, you can make the DNS server accessible only on the Nebula network. Alternatively, listening on `0.0.0.0` will allow anyone that can reach the host to make queries.
 
-See the [serve_dns](#lighthouseserve_dns) docs for more information, or the [example Lighthouse config](#) for syntax.
+The default value for `lighthouse.dns.port` is `53` but you must set an ip address.
```
jasikpark, line 79: Should be one of

```
`dns.port`
`port`
port
```

docs/config/preferred-ranges.mdx
```diff
@@ -0,0 +1,31 @@
...
+Default `preferred_ranges` order: `ipv6` > `public ipv4` > `private ipv4`
+
+There are no magic words for these values:
+
+- `::/0` is all `ipv6`
+- `0.0.0.0/0` is all `ipv4`
+- `10.0.0.0/8` + `172.16.0.0/12` + `192.168.0.0/16` is all `private ipv4` based on [`RFC1918` private ranges](https://datatracker.ietf.org/doc/html/rfc1918)
+- `192.168.0.0/16` would be a specific private `ipv4` network
+
+:::note
```
jasikpark, line 18: This can be moved to the bottom / `has been deprecated in favor of`

jasikpark:
All I have left that I think might happen in this is:

- [ ] Document `lighthouse.advertise_addrs` maybe?
- [ ] Improve `preferred_routes` doc
- [ ] Improve `routines` doc

Alternatively I think it could probably be merged and separate issues / PRs created for these 3

docs/config/lighthouse.mdx
```diff
@@ -60,28 +66,72 @@ $ dig +short @192.168.100.1 192.168.100.5 TXT
...
+The default value for `dns.port` is `53` but you must set an IP address.
 
-See the [serve_dns](#lighthouseserve_dns) docs for more information, or the [example Lighthouse config](#) for syntax.
+See the [`serve_dns`](#lighthouseserve_dns) docs for more information, or the [example Lighthouse config](#) for syntax.
+
+```yml
+dns:
+  # The DNS host defines the IP to bind the dns listener to. This also allows binding to the nebula node IP.
+  host: 0.0.0.0
+  port: 53
```
IanVS, line 87: What do you think about adding `lighthouse:` to the top of each of these example sections?

```suggestion
lighthouse:
  dns:
    # The DNS host defines the IP to bind the dns listener to. This also allows binding to the nebula node IP.
    host: 0.0.0.0
    port: 53
```

docs/config/lighthouse.mdx
```diff
@@ -60,28 +66,72 @@ $ dig +short @192.168.100.1 192.168.100.5 TXT
...
 
 ## lighthouse.dns
 
-dns is used to configure the address (`host`) and port (`port`) the DNS server should listen on. By listening on the host's Nebula IP, you can make the DNS server accessible only on the Nebula network. Alternatively, listening on `0.0.0.0` will allow anyone that can reach the host to make queries.
+`dns` is used to configure the address (`host`) and port (`port`) the DNS server should listen on. By listening on the host's Nebula IP, you can make the DNS server accessible only on the Nebula network. Alternatively, listening on `0.0.0.0` will allow anyone that can reach the host to make queries.
+
+The default value for `dns.port` is `53` but you must set an IP address.
 
-See the [serve_dns](#lighthouseserve_dns) docs for more information, or the [example Lighthouse config](#) for syntax.
+See the [`serve_dns`](#lighthouseserve_dns) docs for more information, or the [example Lighthouse config](#) for syntax.
```
IanVS, line 81: With the example below, I think it's not necessary to point to the example lighthouse config above anymore.

docs/config/lighthouse.mdx
```diff
@@ -60,28 +66,72 @@ $ dig +short @192.168.100.1 192.168.100.5 TXT
...
+```yml
+dns:
+  # The DNS host defines the IP to bind the dns listener to. This also allows binding to the nebula node IP.
+  host: 0.0.0.0
+  port: 53
+```
 
 ## lighthouse.interval
 
+<Pill className="mb-24">Default: 10</Pill>
```
IanVS, line 92: This default doesn't seem to match with the text below.

docs/config/lighthouse.mdx
```diff
@@ -60,28 +66,72 @@ $ dig +short @192.168.100.1 192.168.100.5 TXT
...
+
 `interval` specifies how often a nebula host should report itself to a lighthouse. By default, hosts report themselves to lighthouses once every 60 seconds. Use caution when changing this interval, as it may affect host discovery times in a large nebula network.
 
 ## lighthouse.hosts
 
-**IMPORTANT: THIS SHOULD BE EMPTY ON LIGHTHOUSE NODES**
+:::caution
+
+THIS SHOULD BE EMPTY ON LIGHTHOUSE NODES
```
IanVS, line 100: Let's stop shouting this (lowercase it) now that it's called out in a caution box.

docs/config/listen.mdx
```diff
@@ -17,12 +19,20 @@ listen:
...
 ## listen.port
 
-`port` is the UDP port nebula should use on a host. Setting this to 0 will dynamically assign a port number.
+`port` is the UDP port nebula should use on a host. For a lighthouse node, the port should be defined, conventionally to `4242`, however using port 0 or leaving `port` unset will dynamically assign a port and is recommended for roaming nodes. Using 0 on lighthouses and relay hosts will likely lead to connectivity issues.
+
+## batch
+
+<Pill className="mb-24">Default: 64</Pill>
+
+Sets the max number of packets to pull from the kernel for each syscall (under systems that support `recvmmsg`). Does not support reload
```
IanVS, line 34:
```suggestion
Sets the max number of packets to pull from the kernel for each syscall (on systems that support `recvmmsg`).
```

We don't call out which settings can or cannot be reloaded currently, so I think it's a bit confusing to add the information to this setting.  We should probably find a good way to document all such settings, in a separate PR.

docs/config/logging.mdx
```diff
@@ -11,3 +13,36 @@ logging:
   #disable_timestamp: true
   #timestamp_format: "2006-01-02T15:04:05.000Z07:00"
 ```
+
+## logging.level
+
+<Pill className="mb-24">Default: info</Pill>
+
+options are `panic`, `fatal`, `error`, `warning`, `info`, or `debug`.
```
IanVS, line 21:
```suggestion
Controls the verbosity of logs.  The options are `panic`, `fatal`, `error`, `warning`, `info`, or `debug`.
```

docs/config/logging.mdx
```diff
@@ -11,3 +13,36 @@ logging:
...
+
+## logging.format
+
+<Pill className="mb-24">Default: text</Pill>
+
+options are `json` or `text`
```
IanVS, line 27:
```suggestion
Controls the logging format.  The options are `json` or `text`.
```

docs/config/logging.mdx
```diff
@@ -11,3 +13,36 @@ logging:
...
+
+## logging.disable_timestamp
+
+<Pill className="mb-24">Default: False</Pill>
+
+Disable timestamp logging. Useful when output is redirected to logging system that already adds timestamps.
```
IanVS, line 33:
```suggestion
Disables timestamp logging. Useful when output is redirected to a logging system that already adds timestamps.
```

docs/config/logging.mdx
```diff
@@ -11,3 +13,36 @@ logging:
...
+
+## logging.timestamp_format
+
+`timestamp_format` is specified in Go time format, see:
+https://golang.org/pkg/time/#pkg-constants
```
IanVS, line 38:
```suggestion
https://golang.org/pkg/time/#pkg-constants.
```

docs/config/logging.mdx
```diff
@@ -11,3 +13,36 @@ logging:
...
+
+The default when `format: json` is set is an [RFC3339 timestamp](https://datatracker.ietf.org/doc/html/rfc3339), e.g. `"2006-01-02T15:04:05Z07:00"`.
+
+The default when `format: text` is either seconds since beginning of execution when a TTY is attached, otherwise: `"2006-01-02T15:04:05Z07:00"` (RFC3339).
```
IanVS, line 42:
```suggestion
When `format: text` is set, if a TTY is attached the default is seconds since beginning of execution, otherwise it is RFC3339 (e.g. `"2006-01-02T15:04:05Z07:00"`).
```

docs/config/tun.mdx
```diff
@@ -23,11 +25,15 @@ tun:
...
 Allows the nebula interface (tun) to be disabled, which lets you run a lighthouse without a nebula interface (and therefore without root). You will not be able to communiate over IP with a nebula node that uses this setting.
 
 ## tun.dev
 
-`dev` sets the interface name for your nebula interface.
+`dev` sets the interface name for your nebula interface. If not set, a default will be chosen by the OS.
+For macOS: Not required. If set, must be in the form `utun[0-9]+`.
+For FreeBSD: Required to be set, must be in the form `tun[0-9]+`.
 
 ## tun.drop_local_broadcast
```
IanVS, line 38: If we're specifying defaults, we should probably add the default for `drop_local_broadcast` and `drop_multicast` (I assume they're false, but let's be explicit).

docs/config/relay.mdx
```diff
@@ -16,7 +16,7 @@ relay:
 
 ## relay.relays
 
-Relays are a list of Nebula IP’s that peers can use to relay packets to me. IPs in this list must have am_relay set to true in their configs, otherwise they will reject relay requests.
+`relays` are a list of Nebula IP’s that peers can use to relay packets to me. IPs in this list must have am_relay set to true in their configs, otherwise they will reject relay requests.
```
IanVS, line 19:
```suggestion
`relays` are a list of Nebula IPs that peers can use to relay packets to the host. IPs in this list must have `am_relay` set to `true` in their configs, otherwise they will reject relay requests.
```

docs/config/tun.mdx
```diff
@@ -39,18 +45,42 @@ Toggles forwarding of multicast packets
...
+    - mtu: 8800
+      route: 10.0.0.0/16
+```
+
 ## tun.unsafe_routes
 
-**IMPORTANT NOTE: The nebula certificate of the “via” node MUST have the “route” defined as a subnet in its certificate**
+:::danger
+
+The nebula certificate of the “via” node MUST have the “route” defined as a subnet in its certificate
```
IanVS, line 73:
```suggestion
The nebula certificate of the “via” node MUST have the “route” defined as a subnet in its certificate.
```

docs/config/tun.mdx
```diff
@@ -39,18 +45,42 @@ Toggles forwarding of multicast packets
...
 
 ## tun.mtu
 
+<Pill className="mb-24">Default: 1300</Pill>
+
 Default MTU for every packet, safe setting is (and the default) 1300 for internet routed packets.
 
 ## tun.routes
 
 Route based MTU overrides, you have known vpn ip paths that can support larger MTUs you can increase/decrease them here
```
IanVS, line 60:
```suggestion
Route based MTU overrides. If you have known VPN IP paths that can support larger MTUs you can increase/decrease them here.
```

IanVS (commented):
I like the addition of default pills, more examples, and warning/info banners here.  I had a few grammar / formatting suggestions, but otherwise I think this is looking good.

jasikpark:
Okay, I've pushed a change to improve the `preferred_ranges` doc, including adding a glossary/ underlay network definition

jasikpark:
> Personally I don't think we should have a glossary with just one entry. We only use "underlay" in one other place (the unsafe routes guide), and I think we can either assume folks know what it means or we can add a brief explanation directly where it's used in the preferred_routes doc.
> 
> Another approach would be adding the term to the dn docs glossary and linking to it there, but we don't use the term `underlay` anywhere in those docs or the product, so it's a little confusing too.

I agree having a one-item glossary is a bit wacky -- I figure we might have other places where it makes sense? Or maybe you're right and throwing the definition in the DN glossary makes the most sense :shrug:

I guess a second question is whether we'll reference the DN glossary in this repo or build duplicate pages for the same definitions?

jasikpark:
Would adding more glossary definitions make it better?

docs/config/preferred-ranges.mdx
```diff
@@ -0,0 +1,46 @@
...
+How nebula orders underlay IP addresses it learns about:
+
+Highest priority
+
+- CIDR is in `preferred_ranges` set and is [ipv6](https://en.wikipedia.org/wiki/IPv6)
+- CIDR is in `preferred_ranges` set and is [public ipv4](https://en.wikipedia.org/wiki/IPv4)
+- CIDR is in `preferred_ranges` set and is [private ipv4](https://en.wikipedia.org/wiki/IPv4#Private_networks)
+- CIDR is ipv6
+- CIDR is public ipv4
+- CIDR is private ipv4
```
IanVS, line 24: Since we're talking about how IP addresses are ordered, instead of `CIDR` here (which is a range, not an address), I think we can say `- Address is in `preferred_ranges` and is ipv6`, etc.

jasikpark: I suppose that's fair; there's an implicit noun in these bullet points is currently: "The IP address's CIDR is ipv6"

docs/config/preferred-ranges.mdx
```diff
@@ -16,12 +16,12 @@ How nebula orders underlay IP addresses it learns about:
...
-- CIDR is private ipv4
+- IP address is in `preferred_ranges` set and is [ipv6](https://en.wikipedia.org/wiki/IPv6)
+- IP address is in `preferred_ranges` set and is [public ipv4](https://en.wikipedia.org/wiki/IPv4)
+- IP address is in `preferred_ranges` set and is [private ipv4](https://en.wikipedia.org/wiki/IPv4#Private_networks)
+- IP address is ipv6
+- IP address is public ipv4
+- IP address is private ipv4
 - Route is via a [relay](/config/relay) (experimental)
 
 Lowest priority
```
JohnMaguire, line 27: This heading doesn't quite seem right. "Highest priority" is actually an ordered list of priorities, from highest to lowest. This section seems to talk more about how to manipulate priorities.

jasikpark: good point w/ that - that's not meant to be a heading, but using paragraphs to describe bounds of an unordered list isn't really valid

jasikpark: pushed a diff that should have improved things - I wish there was a clear pattern for labels like I was originally doing, but I think the diff pushed is more correct

jasikpark:
Merged #44 into this, since this isn't as short and sweet as I originally planned. Filled out each config option with docs though they might be pretty sparse.

Maybe I should break this PR out into many individual PRs to give this momentum?

docs/config/cipher.mdx
```diff
@@ -4,8 +4,12 @@ sidebar_position: 6
 
 # cipher
 
-**IMPORTANT: this value must be identical on ALL NODES/LIGHTHOUSES. We do not/will not support use of different ciphers
-simultaneously!**
+:::danger
+
+This value must be identical on ALL NODES/LIGHTHOUSES. We do not/will not support use of different ciphers
```
nbrownus, line 9: > We do not/will not support use

to 

> Nebula does not support use

docs/config/stats.mdx
```diff
@@ -0,0 +1,117 @@
+---
+sidebar_position: 13
+---
+
+import { Pill } from '@components/Pill/Pill';
+
+# stats
+
+Nebula sorts outputting stats for [Graphite](https://graphiteapp.org/) or [Prometheus](https://prometheus.io/). Options
```
nbrownus, line 9: > Nebula sorts outputting stats for

to

> Nebula can provide stats for graphite or prometheus

jasikpark: Oops that was meant to read "supports" - I'll use your wording though thx

nbrownus (approved):

PR #43, opened 2022-08-22
Recommend the LanguageTool VSCode extension for spelling corrections

jasikpark:
This runs via the `languagetools` service you can install on mac via `brew install languagetools` which provides pretty solid grammar/spelling feedback in my experience

https://github.com/davidlday/vscode-languagetool-linter

https://dev.languagetool.org/

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,3 @@
+{
+  "recommendations": ["davidlday.languagetool-linter"]
+}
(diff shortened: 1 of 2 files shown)


IanVS:
Hm, without installing the service, this extension doesn't really do anything, right?  I'm not sure that makes it a good fit for suggesting that anyone who works on this repo should install the extension.  That's not to say we can't use it ourselves, but I am reluctant to suggest that everyone should.

jasikpark:
that's fair, I mostly added it b/c i wanted to add some config derived from the extension into it

IanVS:
I think it's fine to add the settings, but let's not include it in recommended extensions.

jasikpark:
okay 👍

jasikpark:
I'll close this PR and just add the settings while I work on #42 then.

PR #44, opened 2022-08-22
Document `stats` config group

jasikpark:
Broke this out from #42 since it needs a bit more work and information added.

This will add docs for the `stats` config group

### DIFF
--- a/docs/config/stats.mdx
+++ b/docs/config/stats.mdx
@@ -30,9 +30,12 @@ stats:
 
 ## stats.type
 
-Choose your stats provider. Options are `graphite` and `prometheus`. This will determine other available options.
+Choose your stats provider. Options are `graphite` and `prometheus`. If unset, nebula will not report statistics to a provider.
+This will determine other available options.
 
 ## stats.interval
+
+<Pill className="mb-24">Required</Pill>
 
 Duration.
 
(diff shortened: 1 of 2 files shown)


docs/config/stats.mdx
```diff
@@ -54,22 +57,30 @@ Config options if `stats.type` is chosen to be `graphite`
...
-`nebula` by default
+<Pill className="mb-24">
+  Default: <span data-no-transform>nebula</span>
+</Pill>
 
 ### stats.protocol
 
-<Pill className="mb-24">Default: tcp</Pill>
+<Pill className="mb-24">
+  Default: <span data-no-transform>tcp</span>
```
IanVS, line 67: Why use a data-attribute here rather than a className?

jasikpark: eh, not much reason, just seemed like an interesting choice since it's a state of styling w/in the Pill component but I don't want to create a global class / import the Pill component styles?

I don't feel like either is better really 🤷

IanVS: I think if it's really just a flag for styling, a css class name is the best choice.  In my mind, data attributes are really about encoding data into the DOM.

jasikpark: okay! I'd say that data attributes can be interesting to create more complex prop-based styling patterns, but it's not what we normally use here, so I agree that a class is the right choice :p

jasikpark: done

PR #45, opened 2022-08-22
Increase Pill size for better readability

jasikpark:
Currently we have a font-size of `10px` for the Pill component, this sets a font-size of `12px` instead


IanVS (approved):
Approved, but let's get @bgolat's eyes on the pills when he gets back in case he wants to tweak them at all.

jasikpark:
image
bet

PR #56, opened 2022-09-08
Unmark stats.subsystem/stats.namespace as required

JohnMaguire:


docs/config/stats.mdx
```diff
@@ -102,15 +102,19 @@ Path to serve up stats at, for example `/my-stats`
 
 ### stats.namespace
 
-<Pill className="mb-24">Required</Pill>
+<Pill className="mb-24">
+  Default: <span className="no-transform">(none)</span>
+</Pill>
```
IanVS, line 107: For values that don't have a default, I think we normally don't provide a pill at all.  How would you feel about just removing it?  It's a little confusing, IMHO to say that there's a default which is no value.

IanVS: Unless it's literally the value of `(none)`

jasikpark: better to explain what will happen if these aren’t passed

IanVS (approved):

PR #57, opened 2022-09-12
Improve some wording of `stats` config options

jasikpark:


docs/config/stats.mdx
```diff
@@ -92,13 +92,14 @@ Config options if `stats.type` is chosen to be `prometheus`
...
-TCP address of `0.0.0.0:8080` or `:8080` or `192.168.14.12:1234`, for instance.
+Interface for the Prometheus stats server to listen on. TCP address of `0.0.0.0:8080` or `:8080` or
+`192.168.14.12:1234`, for instance.
 
 ### stats.path
 
 <Pill className="mb-24">Required</Pill>
 
-Path to serve up stats at, for example `/my-stats`
+Path to serve up stats at, for example `/my-stats`. Recommended to be set to `/metrics`.
```
IanVS, line 102: Maybe we just change the `for example` to `/metrics`?  I'm not sure we really "recommend" anything, it's just that there's a common default that most will use, right?

IanVS (approved):

docs/config/stats.mdx
jasikpark: sure, I think that's the default in the same sense that `:22` is the default for ssh, right? as in it's what you would use unless you have endpoint conflicts / want security by obscurity

jasikpark: "Path to serve up stats at, most of the time set to `/metrics`." ?

IanVS: How about

>Path to serve stats to, normally `/metrics`.

PR #62, opened 2022-10-21
Add Algolia search

IanVS:
Unfortunately there doesn't seem to be a way to test this other than shipping it to prod.  So, let's :doitlive:!


jasikpark (approved):
lgtm, hopefully it works well

PR #65, opened 2022-11-21
Document nebula CA & signing default validity lengths

### LINKED ISSUE #64: More clearly explain how long certs will last in the quick start
Currently it's not disclosed how long the CA or certificates will be valid for in the nebula quick-start.

We should update the guide to explain this and give hints for how to remember to update your certs before they expire

jasikpark:
Fixes #64 

Another task might be to fully document all possible command line flags for `nebula-cert` and other commands

<https://deploy-preview-65--nebula-docs-dn.netlify.app/docs/guides/quick-start/#creating-your-first-certificate-authority>

This documents that CAs are valid for one year by default, that certificates are valid for one second before their CA expires, and both values can be adjusted by passing `-duration XXhXXmXXs`


IanVS (approved):
looks good to me.

PR #68, opened 2022-12-05
Document DNS + Nebula

### LINKED ISSUE #66: Document how to run nebula w/ public DNS + use the Lighthouse DNS feature
From algolia search analytics, we can see that DNS is useful to be documented (plus I wanna know how it works myself)
image

How do hostnames turn into DNS host names?
Are the names case sensitive?
Do they support Unicode?
How do duplicate hostnames get resolved?
Can I run more than one Lighthouse DNS server?

jasikpark:
Fixes #66 
Fixes [slackhq/nebula/#110](https://github.com/slackhq/nebula/issues/110)

<https://deploy-preview-68--nebula-docs-dn.netlify.app/docs/guides/using-lighthouse-dns/>


jasikpark:
Not totally ready for review, but at a point to where it contains most of the info I know to contain / I need to work on the layout of the document...

docs/guides/using-lighthouse-dns/validateDomain.ts
```diff
@@ -0,0 +1,16 @@
+export const INVALID_DOMAIN = 'Invalid domain';
+
+export function validateDomain(domain: string): true | typeof INVALID_DOMAIN {
+  if (!domain.length) {
+    return true;
+  }
+
+  // sourced from https://www.oreilly.com/library/view/regular-expressions-cookbook/9781449327453/ch08s15.html
+  const domainRegex = new RegExp('\\b((?=[a-z0-9-]{1,63}\\.)(xn--)?[a-z0-9]+(-[a-z0-9]+)*\\.)+[a-z]{2,63}\\b', 'gm');
```
JohnMaguire, line 9:
```suggestion
  const domainRegex = new RegExp('\\b((?=[a-z0-9-]{1,63}\\.)(xn--)?[a-z0-9]+(-[a-z0-9]+)*\\.)+[a-z]{2,63}\\b', 'gmi');
```

This should be a case insensitive match, per the link. Probably worth a test case too.

Perhaps also a test case around international punycode domains too?

docs/guides/using-lighthouse-dns/index.mdx
```diff
@@ -0,0 +1,127 @@
...
+
+## Q&A
+
+- How do hostnames turn into DNS host names?
+  - If the hostname is a valid DNS name, it will be resolved by the Lighthouse DNS resolver.
+- Are the domain names case-sensitive?
+  - Currently the DNS server matches names case sensitively, though DNS is
+    [meant to resolve case-insensitively](https://www.rfc-editor.org/rfc/rfc1035#section-2.3.3).
+- Do they support Unicode?
+  - I don't think so? Maybe via punycode???
```
JohnMaguire, line 101: Who is "they"? We are authoritative here. Want to try creating a cert with Unicode and/or punycode notation and trying it out? I'm guessing we won't properly convert Unicode in a cert to punycode, but the cert name could probably be punycode.

docs/guides/using-lighthouse-dns/index.mdx
```diff
@@ -0,0 +1,127 @@
...
+- How do duplicate hostnames get resolved?
```
JohnMaguire, line 102: Unanswered question here. Give it a try and see what happens? My guess is that there's a race, and you might get either one.

docs/guides/using-lighthouse-dns/index.mdx
```diff
@@ -0,0 +1,127 @@
...
+- Can I run more than one Lighthouse DNS server?
+  - Sure. Just set as many lighthouses as you want in your DNS query settings for your host.
+- How to get the lighthouse to answer questions about itself
```
JohnMaguire, line 105:
```suggestion
- Can the lighthouse resolve its own name?
```

But this is duplicated below.

docs/guides/using-lighthouse-dns/index.mdx
```diff
@@ -0,0 +1,127 @@
...
+  - How does the lighthouse learn about hosts? Hosts connect to the lighthouse as they normally do, and in doing so the
```
JohnMaguire, line 106: Needs to be broken up into two lines? I think this should be unindented too.

docs/guides/using-lighthouse-dns/index.mdx
```diff
@@ -0,0 +1,127 @@
...
+:::
+
+Nebula comes with built-in DNS server support in the Lighthouse hosts.
+
+Lighthouse DNS can be helpful and useful for generating DNS records based on dynamic nebula hosts, for example if you
+are spinning up new nebula hosts with demand, or using [anycast](https://en.wikipedia.org/wiki/Anycast) addresses to
+split load.
+
+An alternative approach to DNS would be to point public DNS records at private nebula IP addresses. If you want to, you
+can simply create a public DNS A record pointing to `10.0.0.23` for your private team wiki on <https://gandi.net> or
```
JohnMaguire, line 16: Why call out gandi.net here? We do call out DigitalOcean elsewhere and they offer free DNS, as does Cloudflare, but I'm not sure we need to call anything out.

docs/guides/using-lighthouse-dns/index.mdx
```diff
@@ -0,0 +1,127 @@
...
+some other public DNS, set up a server w/ TLS and Basic Auth via
+[caddy server](https://caddyserver.com/docs/getting-started), and you're golden.
```
JohnMaguire, line 18: Also, why call out Caddy specifically? Any web server should work. I know Caddy has nice built-in Let's Encrypt support but that won't work with a Nebula host anyway.

I'm not sure what the purpose of Basic Auth is here either. Nebula is an authentication layer already (granted w/o authorization, but that should probably be done at the app layer anyway, not the web server layer either.) Finally, TLS isn't strictly necessary either (though it's a good idea!) and might be a distraction from the DNS part of this guide.

docs/guides/using-lighthouse-dns/index.mdx
```diff
@@ -0,0 +1,127 @@
...
+
+To get started, set up a lighthouse and two hosts using our [Quick Start](/docs/guides/quick-start/).
+
+You can then use [`lighthouse.serve_dns`](/docs/config/lighthouse#lighthouseserve_dns) and the
+[`lighthouse.dns`](/docs/config/lighthouse#lighthousedns) config settings for your lighthouse config file to enable DNS
+querying.
```
JohnMaguire, line 24: To be pedantic, it's enabling the DNS server. Querying is always possible! :P

docs/guides/using-lighthouse-dns/index.mdx
```diff
@@ -0,0 +1,127 @@
...
+
+:::note
+
+Only Lighthouses should have `lighthouse.serve_dns` enabled, as DNS info is aggregated as hosts talk to the lighthouse,
+and if enabled on a regular host would only respond with answers for hosts it has already handshaked with
```
JohnMaguire, line 29:
```suggestion
and if enabled on a regular host would only respond with answers for hosts it has already handshaked with.
```

But actually, more to the point, Nebula will refuse to serve DNS on a non-Lighthouse:

```
if c.GetBool("lighthouse.serve_dns", false) {
    if c.GetBool("lighthouse.am_lighthouse", false) {
        serveDns = true
    } else {
        l.Warn("DNS server refusing to run because this host is not a lighthouse.")
    }
}
```

docs/guides/using-lighthouse-dns/index.mdx
```diff
@@ -0,0 +1,127 @@
...
+
+:::
+
+You'll then want to set up the lighthouse as a DNS server for the other two hosts. This can be either the public static
+lighthouse IP or the private Nebula IP. If you set `lighthouse.dns.host: [::]`, it will bind to all interfaces,
```
JohnMaguire, line 34:
```suggestion
lighthouse IP or the private Nebula IP depending on the Lighthouse's configuration. By setting `lighthouse.dns.host` to `[::]`, nebula will bind to all interfaces,
```

Also I'm noticing nebula is sometimes capitalized and sometimes not in this file. Do we have a style guide around this?

docs/guides/using-lighthouse-dns/index.mdx
```diff
@@ -0,0 +1,127 @@
...
+including both the public and Nebula IP. Binding to only the Nebula IP, for example `lighthouse.dns.host: 10.0.0.1` will
+ensure the DNS is only accessible to hosts that are allowed to make UDP requests to that lighthouse.
```
JohnMaguire, line 36:
```suggestion
ensure the DNS is only accessible to hosts that are allowed to contact the lighthouse via nebula.
```

Whether other hosts can hit it via UDP is a different question.

docs/guides/using-lighthouse-dns/index.mdx
```diff
@@ -0,0 +1,127 @@
...
+
+```yaml
+lighthouse:
+  serve_dns: true
+  host: [::]
+  port: 53
+```
+
+to your config for your lighthouse as well as a firewall rule that enables UDP on port `53` for all hosts you want to
+query DNS on the lighthouse.
```
JohnMaguire, line 50:
```suggestion
be able to query DNS on the lighthouse.
```

docs/guides/using-lighthouse-dns/validateDomain.ts
jasikpark: the test should be covered by "should validate generated domains" but I'll enter an explicit test as well. I assume that this regex is already case insensitive?

docs/guides/using-lighthouse-dns/index.mdx
```diff
@@ -0,0 +1,127 @@
...
+
+```bash
+dig @100.100.0.1 +short "100.100.0.2" TXT
+"Name: alice-laptop" "Ips: [100.100.0.2/22]" "Subnets []" "Groups [role:Laptop]" "NotBefore 2022-12-06 21:34:25 +0000 UTC" "NotAFter 2024-05-13 17:37:27 +0000 UTC" "PublicKey 2c4828672fef1306124df94eb0fecd753e462e1bd4107f866f4e1a463550eb1b" "IsCA false" "Issuer 5b568aedaa5d97746e870f01a356ba474cf3251c0e743499ec668f93efa77f51"
+```
+
+> The IP is the closest thing to a nebula "primary key" of sorts. An overlapping name/etc. is probably a mistake that
+> can happen commonly, but only one host can actively have an IP (from the lighthouse perspective) at any moment. (note:
+> if you give the same IP to two hosts, via valid certs, they will fight back and forth, but at the moment you query it
+> will be correct.)
```
JohnMaguire, line 72: > The IP is the closest thing to a nebula "primary key" of sorts. An overlapping name/etc. is probably a mistake that
> can happen commonly, but only one host can actively have an IP (from the lighthouse perspective) at any moment. (note:
> if you give the same IP to two hosts, via valid certs, they will fight back and forth, but at the moment you query it
> will be correct.)

I don't think we need to go into a ton of detail about what happens in the DNS server if you have host's with conflicting IP addresses. This is a misconfiguration that will cause general connectivity issues besides DNS. We can discuss what happens with overlapping names more clearly perhaps though? I think the following captures it but I haven't tested the behavior...

> In Nebula, a node's IP address is the only identifier that MUST be globally unique. Therefore, it's possible to have two hosts with the same certificate name, but different IP addresses. Unfortunately, this can cause inconsistent results when querying for these hosts' hostname as the DNS query is ambiguous.

docs/guides/using-lighthouse-dns/validateDomain.ts
IanVS: It's only using lowercase letters, so no it's not.  The source you linked to specifies:

```
Regex options: Case insensitive
```

docs/guides/using-lighthouse-dns/index.mdx
```diff
@@ -0,0 +1,127 @@
...
+
+If I then host a server on `[::]:3000` on `alice-laptop` and set up a firewall rule that allows this to be accessed from
```
JohnMaguire, line 74: Maybe stick to port `:80` and remove the port from cURL for simplicity.

docs/guides/using-lighthouse-dns/index.mdx
```diff
@@ -0,0 +1,127 @@
...
+`bob-laptop`, I can request it with `curl` using the custom DNS resolver.
```
JohnMaguire, line 75:
```suggestion
`bob-laptop`, I can request it with `curl` using the custom DNS resolver. (In a production environment, you may choose to configure the system resolver.)
```

JohnMaguire:
This looks pretty good. It's nice that it walks through how to set up the DNS server, its caveats, and how to query it. It's too bad we don't have better resources to point to for how to configure system DNS, but this is probably highly specific to each environment (e.g. configuring it via provisioning tools vs. manual config on each distro, etc.) and should be left out, at least for now.

jasikpark:
> how to configure system DNS, but this is probably highly specific to each environment (e.g. configuring it via  provisioning tools vs. manual confi
this is also hard to do since the lighthouse DNS does not resolve queries recursively and only responds about nebula ips / hosts

JohnMaguire:
> > how to configure system DNS, but this is probably highly specific to each environment (e.g. configuring it via  provisioning tools vs. manual confi
> > this is also hard to do since the lighthouse DNS does not resolve queries recursively and only responds about nebula ips / hosts

Right, you'd need a tool like dnsmasq or something probably.

docs/guides/using-lighthouse-dns/index.mdx
jasikpark: Eh, :3000 is the default for lots of localhost servers for me.. I figure it's useful to have the port specified && if they know about :80 they can leave off the flag 🤷

docs/guides/using-lighthouse-dns/index.mdx
jasikpark: I tried with a host named `caleb-dns-test-xn--5l8h-ubuntu` and it didn't seem to grab it for some reason, even though I was communicating with the DNS server on the lighthouse from that host via Nebula, so it definitely had handshaked

jasikpark:
```bash
dig @100.100.0.20 +short caleb-dns-test-xn--5l8h-ubuntu A
100.100.0.28
```

oh wait. success! I'll document this w/ an example

docs/guides/using-lighthouse-dns/index.mdx
jasikpark: Specifically recommending adding auth to services & not just relying on nebula == auth 🤷 

I'll pull out the specific services mentioned though, just added those for the draft

docs/guides/using-lighthouse-dns/index.mdx
jasikpark: In dn-docs, we have a style of not capitalizing "nouns" but I'm wont to do so 😓

docs/guides/using-lighthouse-dns/index.mdx
jasikpark: Eh, I suppose you could technically query over TCP DNS :p

jasikpark:
I feel like providing that `dnsmasq` guide would be helpful if easily implemented

docs/guides/using-lighthouse-dns/index.mdx
JohnMaguire: My point is that UDP packets can be sent and received without Nebula. Setting the host to the Nebula IP restricts it to packets that come in over Nebula, not packets that are UDP.

docs/guides/using-lighthouse-dns/index.mdx
```diff
@@ -0,0 +1,142 @@
...
+
+Nebula comes with built-in DNS server support in the Lighthouse hosts.
+
+Lighthouse DNS can be helpful and useful for generating DNS records based on dynamic nebula hosts, for example if you
+are spinning up new nebula hosts with demand, or using [anycast](https://en.wikipedia.org/wiki/Anycast) addresses to
+split load.
+
+An alternative approach to DNS would be to point public DNS records at private nebula IP addresses. If you want to, you
+can simply create a public DNS A record pointing to `10.0.0.23` for your private team wiki on a public nameserver, set
+up a server w/ TLS and Basic Auth, and you're golden.
```
JohnMaguire, line 17: I still think that TLS and Basic Auth are irrelevant to a guide about Nebula.

If the goal for including these here is an attempt to note that Nebula shouldn't be used for app-level authorization/authentication I think you should do that explicitly. As it stands, it's not clear to me what you're trying to say here.

I think the vast majority of internal apps either (a) already have authentication built into them, or (b) don't require it.

I'd probably just leave the whole discussion out of this guide for simplicity.

> An alternative approach to DNS would be to point public DNS records at private nebula IP addresses. For example, you can create a public DNS A record pointing to `10.0.0.23` for your private team wiki. While this does make your internal IP address visible to the internet, only users on your Nebula network will be able to access it.

docs/guides/using-lighthouse-dns/index.mdx
```diff
@@ -0,0 +1,142 @@
...
+
+Lighthouse DNS in Nebula is Experimental and not intended to be a robust solution currently.
+
+:::
+
+Nebula comes with built-in DNS server support in the Lighthouse hosts.
+
+Lighthouse DNS can be helpful and useful for generating DNS records based on dynamic nebula hosts, for example if you
+are spinning up new nebula hosts with demand, or using [anycast](https://en.wikipedia.org/wiki/Anycast) addresses to
+split load.
```
JohnMaguire, line 13: How does DNS relate to anycast? I'm not very familiar with it.

docs/guides/using-lighthouse-dns/index.mdx
```diff
@@ -0,0 +1,142 @@
...
+
+:::caution
+
+Lighthouse DNS in Nebula is Experimental and not intended to be a robust solution currently.
+
+:::
+
+Nebula comes with built-in DNS server support in the Lighthouse hosts.
+
+Lighthouse DNS can be helpful and useful for generating DNS records based on dynamic nebula hosts, for example if you
```
JohnMaguire, line 11: nit: maybe just "helpful" or "useful"? These seem to be synonyms.

docs/guides/using-lighthouse-dns/index.mdx
```diff
@@ -0,0 +1,142 @@
...
+are spinning up new nebula hosts with demand, or using [anycast](https://en.wikipedia.org/wiki/Anycast) addresses to
+split load.
+
+An alternative approach to DNS would be to point public DNS records at private nebula IP addresses. If you want to, you
+can simply create a public DNS A record pointing to `10.0.0.23` for your private team wiki on a public nameserver, set
+up a server w/ TLS and Basic Auth, and you're golden.
+
+To get started, set up a lighthouse and two hosts using our [Quick Start](/docs/guides/quick-start/).
```
JohnMaguire, line 19:
```suggestion
To get started, set up a lighthouse and two hosts using our [Quick Start guide](/docs/guides/quick-start/).
```

docs/guides/using-lighthouse-dns/index.mdx
```diff
@@ -0,0 +1,142 @@
...
+
+You can then use [`lighthouse.serve_dns`](/docs/config/lighthouse#lighthouseserve_dns) and the
+[`lighthouse.dns`](/docs/config/lighthouse#lighthousedns) config settings for your lighthouse config file to enable DNS
+answers served by Lighthouse DNS.
```
JohnMaguire, line 23: This end reads a little awkward to me. Maybe just "to enable the (built-in?) DNS server" or even "to enable Lighthouse DNS"? I've not heard us refer to it as "Lighthouse DNS" anywhere else though.

docs/guides/using-lighthouse-dns/index.mdx
```diff
@@ -0,0 +1,142 @@
...
+
+:::note
+
+Only Lighthouses should have `lighthouse.serve_dns` enabled, as DNS info is aggregated as hosts talk to the lighthouse,
```
JohnMaguire, line 27:
```suggestion
Only Lighthouses should have `lighthouse.serve_dns` enabled, as DNS info is collected when hosts report to the lighthouse.
```

docs/guides/using-lighthouse-dns/index.mdx
```diff
@@ -0,0 +1,142 @@
...
+and if enabled on a regular host, nebula will warn: `DNS server refusing to run because this host is not a lighthouse.`
```
JohnMaguire, line 28:
```suggestion
If enabled on a regular host, nebula will warn: `DNS server refusing to run because this host is not a lighthouse.`
```

docs/guides/using-lighthouse-dns/index.mdx
```diff
@@ -0,0 +1,142 @@
...
+and refuse to serve DNS on that non-lighthouse host.
```
JohnMaguire, line 29: probably redundant based on the nebula warning text. I'd pick one or the other. I'd probably choose this one actually since the warning log line could change in the future and its exact text is not necessary here.

> nebula will not honor the option if enabled on a non-lighthouse host.

docs/guides/using-lighthouse-dns/index.mdx
```diff
@@ -0,0 +1,142 @@
...
+
+:::
+
+You'll then want to set up the lighthouse as a DNS server for the other two hosts. This can be either the public static
+lighthouse IP or the private Nebula IP depending on the Lighthouse's configuration. By setting `lighthouse.dns.host` to `[::]`, nebula will bind to all interfaces,
+including both the public and Nebula IP. Binding to only the Nebula IP, for example `lighthouse.dns.host: 10.0.0.1` will
+ensure the DNS is only accessible to hosts that are allowed to contact the lighthouse via nebula.
```
JohnMaguire: This paragraph is a little confusing because it starts by telling the user they should modify their non-lighthouse hosts, but then goes on to talk about a configuration setting on the lighthouse host. Perhaps we should discuss `lighthouse.dns.host` prior to this paragraph and then tell them to configure the non-lighthouse hosts?

docs/guides/using-lighthouse-dns/index.mdx
JohnMaguire: I think that's a Javascript thing. The default HTTP port is 80. Since this guide is about DNS I'd avoid getting into the weeds on other systems. This is because when writing technically, I try to reduce any unrelated noise if possible. This is super minor and 1/5 GAS though.

docs/guides/using-lighthouse-dns/index.mdx
```diff
@@ -0,0 +1,142 @@
...
+curl --dns-servers "100.100.0.1" http://alice-laptop:3000
+<div>hello i am a website</div>
+```
+
+## Name hosts with valid domain names
+
+Hosts with names conforming to the [DNS RFC](https://www.rfc-editor.org/rfc/rfc1035#section-2.3.1) will be resolved,
+otherwise the Lighthouse DNS will return an empty response.
+
+Here's a validator inline in the documentation for you:
```
JohnMaguire:
```suggestion
You can use the following tool to check if a hostname is valid:
```

docs/guides/using-lighthouse-dns/index.mdx
```diff
@@ -0,0 +1,142 @@
...
+    ```bash
+    # `alice-xn--3s8h` = `alice-💻` in punycode.
+    dig @100.100.0.1 +short alice-xn--3s8h A
+    100.100.0.2
+    ```
+    Nebula supports full unicode strings in the host name, but for hosts to be discovered via Lighthouse DNS, they must
+    conform to the DNS spec. <https://www.punycoder.com/> is a useful tool for conversion.
+- Are the domain names case-sensitive?
+  - Currently the DNS server matches names case sensitively, though DNS is
+    [meant to resolve case-insensitively](https://www.rfc-editor.org/rfc/rfc1035#section-2.3.3).
```
JohnMaguire, line 104: There's an open bug in the nebula tracker you may consider linking here: https://github.com/slackhq/nebula/issues/776

docs/guides/using-lighthouse-dns/index.mdx
```diff
@@ -0,0 +1,142 @@
...
+- How do duplicate hostnames get resolved?
+  - In Nebula, a node's IP address is the only identifier that MUST be globally unique. Therefore, it's possible to have
+    two hosts with the same certificate name, but different IP addresses. Unfortunately, this can cause inconsistent
+    results when querying for these hosts' hostname as the DNS query is ambiguous. The lighthouse will return the nebula
+    IP of whichever host it has most recently performed a handshake with.
+- Can I run more than one Lighthouse DNS server?
+  - Sure. Just set as many lighthouses as you want in your DNS query settings for your host.
```
JohnMaguire, line 111: I'm not sure this answer makes sense to me. I'd maybe just leave this question off?

JohnMaguire (commented):
Woops, didn't realize this didn't submit earlier.

jasikpark:
lol

docs/guides/using-lighthouse-dns/index.mdx
```diff
@@ -0,0 +1,138 @@
...
+>
+> `Add-DnsClientNrptRule -Namespace ".<domain>" -NameServers "<lighthouse ip>"` using powershell will tell windows to
+> include your lighthouse dns server in name lookups.
+>
+> This means that your nebula hosts will need to share a domain name. You may be able to omit the -Namespace option but
+> I didn't try it.
+>
+> https://github.com/slackhq/nebula/issues/110#issuecomment-578470077
+
+Explain how to set up groups for firewall rules?
```
JohnMaguire, line 130: FWIW, I think groups and firewall rules can be a separate article. We can do an "any" rule for the firewall rule example.

docs/guides/using-lighthouse-dns/index.mdx
```diff
@@ -0,0 +1,138 @@
...
+
+Document how to set up DNS consumption w/ `dnsmasq` on Ubuntu?
```
JohnMaguire, line 132: I would avoid this, at least for this initial publish. dnsmasq is tricky and there are probably security questions we don't want to wade into here.

docs/guides/using-lighthouse-dns/index.mdx
```diff
@@ -0,0 +1,138 @@
...
+
+see <https://github.com/DefinedNet/nebula-docs/pull/68#issuecomment-1343201269> and these slackhq/nebula issues:
+
+- https://github.com/slackhq/nebula/issues/434
+- https://github.com/slackhq/nebula/issues/110
```
JohnMaguire, line 137: I think we can close this one out when this lands.

docs/guides/using-lighthouse-dns/index.mdx
```diff
@@ -0,0 +1,138 @@
...
+- How does the lighthouse learn about hosts?
+  - Hosts connect to the lighthouse as they normally do, and in doing so the lighthouse learns about their cert. Due to
+    this fact, the lighthouse can only answer questions about hosts it has seen since last start.
+- Can the lighthouse resolve its own name?
+  - As of nebula `v1.6.1`, no, the lighthouse only responds about hosts it has handshaked with, it never handshakes with
+    itself. [slackhq/nebula/issues/560](https://github.com/slackhq/nebula/issues/560) is tracking this feature.
+
+## TODO
+
+Document Windows DNS setup?
```
JohnMaguire, line 118: Meh. Unless we can confirm this 100% works as expected I might punt on this one. If we lay out the command you listed below we may get support questions about it in Nebula OSS Slack, and I'm not prepared to answer them (personally.)

JohnMaguire (commented):
Some thoughts on the TODOs. I'm excited to get this posted!

docs/guides/using-lighthouse-dns/index.mdx
```diff
@@ -45,8 +45,17 @@ lighthouse:
   port: 53
 ```
 
-to your config for your lighthouse as well as a firewall rule that enables UDP on port `53` for all hosts you want to be
-able to query DNS on the lighthouse.
+to your config for your lighthouse as well as a [firewall rule](/docs/config/firewall) in your lighthouse config that
+enables UDP on port `53` for all hosts you want to be able to query DNS on the lighthouse. For example to allow only
+hosts with the `role:Laptop` group:
```
JohnMaguire: I think we should just leave this as any. I don't see a reason to introduce the concept of groups, (or even more confusingly, "role:") here.

jasikpark: I'll do so, I just don't like documenting worse practices just b/c they're less complicated 🤷

jasikpark: I suppose your point is that suggesting a group named `role:Laptop` is worse than none at all?

JohnMaguire: @jasikpark Yeah, there's a few things here. "role:" is an artifact of DN, and using that term here, when talking about groups, confuses things. This is Nebula docs, not DN docs. In Nebula, you establish groups. You could take a page out of DN's book and use a "role:" prefix, but basically this gets into a whole discussion on naming groups that we don't need to have on this page.

Furthermore, a role named "laptop" doesn't get at the core of how we expect firewall rules to be used (or groups for that matter.) It typically makes more sense to have groups/roles that dictate the purpose of a host, like "db" or "api-db" and "web" or "api-web" so that you can make rules around who should have access to what, based on their needs.

But with the DNS system, I expect most users want to expose it to their entire network.

jasikpark: thanks for that explanation; & yeah, I wasn't thinking that DNS is probably a universal need on the network

JohnMaguire (approved):

PR #72, opened 2022-12-07
Add nebula docs source to footer

jasikpark:
We only link to slackhq/nebula currently

### DIFF
--- a/docusaurus.config.cjs
+++ b/docusaurus.config.cjs
@@ -89,7 +89,7 @@ const config = {
               },
               {
                 label: 'Docs Github',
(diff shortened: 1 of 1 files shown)


docusaurus.config.cjs
```diff
@@ -87,6 +87,10 @@ const config = {
                 label: 'Config Reference',
                 to: '/docs/config',
               },
+              {
+                label: 'Docs Github',
+                to: 'https://github.com/DefinedNet/nebula-docs',
```
IanVS, line 92: This should be href, since it's an external link.

PR #76, opened 2022-12-09
Add note about blocklist distribution

JohnMaguire:
This closes out the documentation task https://github.com/slackhq/nebula/issues/304.

### DIFF
--- a/docs/config/pki.mdx
+++ b/docs/config/pki.mdx
@@ -54,9 +54,16 @@ hosts.
 
 ## pki.blocklist
 
-The `blocklist` contains a list of individual host certificates to ignore. In the case a host’s credentials are stolen
-or compromised, this allows us to block connectivity from a host, even if it is signed by a certificate authority we
-trust.
+:::note
+
+The blocklist is _not_ distributed via Lighthouses. To ensure access to your entire network is blocked you must distribute
+the full blocklist to every host in your network. This is typically done via tooling such as Ansible, Chef, or Puppet.
+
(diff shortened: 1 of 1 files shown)


docs/config/pki.mdx
```diff
@@ -54,9 +54,16 @@ hosts.
 
 ## pki.blocklist
 
-The `blocklist` contains a list of individual host certificates to ignore. In the case a host’s credentials are stolen
-or compromised, this allows us to block connectivity from a host, even if it is signed by a certificate authority we
-trust.
+:::note
+
+The blocklist is _not_ distributed via Lighthouses. To ensure access to your entire network is blocked you must distribute
```
jasikpark, line 59: maybe be more explicit about it being a static config value to update?

jasikpark (approved):

docs/config/pki.mdx
JohnMaguire: I'm not sure I understand. What do you mean by static config value to update?

jasikpark: "you must distribute" = "update the configuration file on each host & HUP them" I guess I was arguing be more explicit

JohnMaguire: Let's leave it like this for now and add a separate section on updating configs (including how to reload them.) Then we can link to it from here if we think it's helpful. But in general I'm trying to focus on a specific config item (`pki.blocklist`) and not get into the weeds on how the config itself works. :)

PR #83, opened 2022-12-13
Various edits and tweaks for clarity

JohnMaguire:
Went around trying to do some cleanup...

- My `:::note` in "Sign Certificates with Public Keys" guide is really a warning, so I switched it to `:::caution`. It could even be considered danger, but I want to avoid overuse of red and we use private key copying in examples elsewhere.
- I changed a `:::danger` in unsafe_routes docs to `:::caution`. I think the worst that happens here is you tear your hair out a little bit during setup, so I wouldn't consider this dangerous per se.
- "EXPERIMENTAL: asdf" looked a little jarring to me on the relays page, so I reworded it. I also want to encourage people to use it as it's a very useful tool. I changed it from `:::note` to `:::caution` because we use caution for other experimental notes (e.g. DNS.)
- Clarified ciphers used and recommended AES.
- Explained how `relays.relays` works a little more as it's "backwards" of what most seem to expect.
- Moved things around a little bit in the DNS guide to try to get to the meat of it quicker, and moved limitations further down the page (also fleshed them out a bit more.) The DNS hostname validity tool now lives under the Limitations section.

### DIFF
--- a/docs/config/cipher.mdx
+++ b/docs/config/cipher.mdx
@@ -6,13 +6,16 @@ sidebar_position: 6
 
 :::danger
 
-This value must be identical on ALL NODES/LIGHTHOUSES. Nebula does not support use of different ciphers simultaneously!
+This value must be identical on ALL nodes and lighthouses. Nebula does not support the use of different ciphers
+simultaneously!
 
 :::
 
-`cipher` allows you to choose between the available ciphers for your network. The default is `aes`, and the available
-options are `chachapoly` or `aes`.
+`cipher` allows you to choose between the available ciphers for your network. You may choose `chachapoly` to use the
+ChaCha20-Poly1305 cipher or `aes` for AES256-GCM.
+
+Because most devices have hardware support for the AES instruction set (e.g. AES-NI), this is the recommended option.
 
 ```yml
-cipher: chachapoly
+cipher: aes
 ```
--- a/docs/config/relay.mdx
+++ b/docs/config/relay.mdx
@@ -4,9 +4,10 @@ sidebar_position: 8
 
 # relay
 
-:::info
+:::caution
 
-EXPERIMENTAL: relay support for networks that can’t establish direct connections.
+Relay support is new and therefore should be considered experimental. However, relays are a useful tool for solving
+tricky connectivity issues between hosts that otherwise are unable to route packets to each other.
 
 :::
 
@@ -20,14 +21,17 @@ relay:
 
 ## relay.relays
 
-`relays` are a list of Nebula IPs that peers can use to relay packets to the host. IPs in this list must have `am_relay`
-set to `true` in their configs, otherwise they will reject relay requests.
+`relays` are a list of Nebula IPs that peers can use to relay packets to this host. IPs in this list must have
+`am_relay` set to `true` in their configs, otherwise they will reject relay requests.
 
 ```yml
 relays:
   - 192.168.100.1
   - <other Nebula VPN IPs of hosts used as relays to access me>
 ```
+
+This list of relays is reported to the Lighthouse. When other nodes attempt to handshake with this host, the Lighthouse
+will indicate its supported relays in addition to its known IP addresses.
 
 ## relay.am_relay
 
--- a/docs/guides/using-lighthouse-dns/index.mdx
+++ b/docs/guides/using-lighthouse-dns/index.mdx
@@ -2,7 +2,8 @@
 
 :::caution
 
-Lighthouse DNS in nebula is Experimental and not intended to be a robust solution currently.
+Lighthouse DNS in nebula is experimental and should not be considered to be a robust solution. For details, see the
+[limitations](#limitations) listed below.
 
 :::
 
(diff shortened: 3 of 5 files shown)


docs/config/cipher.mdx
```diff
@@ -6,13 +6,16 @@ sidebar_position: 6
 
 :::danger
 
-This value must be identical on ALL NODES/LIGHTHOUSES. Nebula does not support use of different ciphers simultaneously!
+This value must be identical on ALL nodes and lighthouses. Nebula does not support the use of different ciphers
```
jasikpark, line 9: may be useful to document what happens if you use different ciphers; ig just non-working network

docs/config/relay.mdx
```diff
@@ -4,9 +4,10 @@ sidebar_position: 8
 
 # relay
 
-:::info
+:::caution
 
-EXPERIMENTAL: relay support for networks that can’t establish direct connections.
+Relay support is new and therefore should be considered experimental. However, relays are a useful tool for solving
```
jasikpark, line 9: This was added when relays were, so it makes sense to reword this now for sure

docs/guides/using-lighthouse-dns/index.mdx
```diff
@@ -43,12 +32,19 @@ lighthouse:
     port: 53
 ```
 
-to your config for your lighthouse. By setting `lighthouse.dns.host` to `[::]`, nebula will bind to all interfaces,
```
jasikpark, line 46: I would keep the comma, it shows the completion of a clause - "nebula with bind to all interfaces. this includes the public IP and the nebula IP"

docs/guides/using-lighthouse-dns/index.mdx
```diff
@@ -86,15 +84,29 @@ curl --dns-servers "100.100.0.1" http://alice-laptop:3000
...
+- The built-in DNS server will only respond to queries for known Nebula hosts and will not make upstream queries to a
+  secondary DNS server (e.g. for supporting both internal DNS and public DNS.)
+- A host's hostname is restricted to the name in its certificate.
+- There is no way to register additional hostnames (e.g. additional CNAMEs for a given Nebula host, or additional A
+  records for managing non-Nebula hosts.)
+- Duplicate names in certificates will result in non-stable answers from the DNS server.
+- If the name in the Nebula certificate is not a [valid hostname](https://www.rfc-editor.org/rfc/rfc1035#section-2.3.1),
+Lighthouse DNS will return an empty result.
+
+### Hostname Validity
```
jasikpark, line 98: Why is this an `h3`? It's not a subsection of Limitations I think?

jasikpark (approved):
looks great, cleans up a good bit

docs/config/cipher.mdx
JohnMaguire: Actually I did at first and then decided it wasn't helpful. Basically the gist is "nodes with different ciphers won't be able to handshake." I think it's kind of self-evident, and it can be hard to explain specifics (what if the lighthouse & one node are aes, and another node is chacha. It'll show up differently on different nodes.)

docs/guides/using-lighthouse-dns/index.mdx
JohnMaguire: Per the PR description I made it a sub-section of limitations. My understanding is that we have a hostname validity checker because there's a limitation that names in the certificate must be valid hostnames for DNS to work. You can see that's my last bullet point just above here.

docs/guides/using-lighthouse-dns/index.mdx
JohnMaguire: https://www.brandeis.edu/writing-program/resources/students/academic/style-grammar/comma-rules.html

I think the comma exists to suffix the introductory clause (point 2). I think the previous usage was an attempt at point 3? When you use rule 3 from this list you should be able to drop the center clause and have the sentence still make sense.

> By setting `lighthouse.dns.host` to `[::]` including both the public and nebula IP.

1/5 GAS though, if you make a review suggestion I'll commit it! :)

docs/guides/using-lighthouse-dns/index.mdx
jasikpark: Eh, I guess you could call that a limitation, I still feel like it's a larger part of the feature inherent in the feature; "a limitation of DNS over Lighthouse is that it is constrained by DNS RFCs"

ig that's a limitation all the same

JohnMaguire: OK! Do you just want me to make this an h2 again? Or move it?

JohnMaguire: I'm not sure where it makes the most sense to live.

jasikpark: Maybe pull the Q&A section out into that "Hostname Validity" section? Can deff consolidate there... could also mention what happens w/ duplicate names in the section too? or not

jasikpark (approved):

PR #87, opened 2022-12-21
Validate hostnames instead of domain names

jasikpark:
Changes:
- updates the hostname validator to accept hostnames in addition to domain names
- remaps the input to lower case to indicate the case insensitive DNS matching
- uses `parse-domain` to provide useful errors
- `npm run format`

test w/ https://deploy-preview-87--nebula-docs-dn.netlify.app/docs/guides/using-lighthouse-dns/#hostname-validity


jasikpark:
I should duplicate
```

/* Prevent ios from "zooming in" */
.navbar__search-input {
  font-size: 1rem;
}

```

docs/guides/using-lighthouse-dns/validateHostname.ts
```diff
@@ -0,0 +1,20 @@
+import { parseDomain, ParseResultType } from 'parse-domain';
```
JohnMaguire, line 1: parse-domain seems primarily concerned with parsing a domain using the public suffix list. Its package notes:

> 💡 Please note: [publicsuffix.org](https://publicsuffix.org/) is updated several times per month. This package comes with a prebuilt list that has been downloaded at the time of npm publish. In order to get an up-to-date list, you should run npx parse-domain-update everytime you start or build your application. This will download the latest list from https://publicsuffix.org/list/public_suffix_list.dat.

If we don't set this hook up, someone else may rely on the package without realizing it's not fully configured. Should we configure this?

docs/config/pki.mdx
```diff
@@ -56,8 +56,9 @@ hosts.
 
 :::note
 
-The blocklist is _not_ distributed via Lighthouses. To ensure access to your entire network is blocked you must distribute
-the full blocklist to every host in your network. This is typically done via tooling such as Ansible, Chef, or Puppet.
+The blocklist is _not_ distributed via Lighthouses. To ensure access to your entire network is blocked you must
+distribute the full blocklist to every host in your network. This is typically done via tooling such as Ansible, Chef,
+or Puppet.
```
JohnMaguire, line 61: Why did so many lines get updated for line length? Did prettier change?

JohnMaguire (approved):

docs/guides/using-lighthouse-dns/validateHostname.ts
jasikpark: image
we should be fine, since I'm treating the last two cases as the same?

docs/config/pki.mdx
jasikpark: I dunno.. I wonder if mdx is being ignored rn?

docs/guides/using-lighthouse-dns/validateHostname.ts
JohnMaguire: > we should be fine, since I'm treating the last two cases as the same? 

By "someone else may rely on this package without realizing it's not fully configured" I meant that even though we aren't using the PSL behavior today we _might_ want to use it in the future for something. Seeing that the dependency is already installed in the project, will we realize it's not fully configured, per the library's documentation?

Is there a reason to use this library instead of simplifying the original regex solution? Many steer away from regex, and it can certainly be difficult to understand and modify - and at times can cause performance concerns. In this case though, the original regex looked fairly simple and we might be able to tweak it to be even simpler (or find another one we can borrow from the internet.)

The reason I bring up this alternative is that it might offer a way to reduce our dependency on third-party code that is unnecessarily complex for our use case.

jasikpark: Fair point, I appreciated this for classifying inputs and telling you what characters are invalid rather than a broad yes/no answer

JohnMaguire (approved):

docs/config/pki.mdx
JohnMaguire: @jasikpark Should we file a task to investigate?

PR #89, opened 2023-01-13
Add CI

jasikpark:
Fixes https://github.com/DefinedNet/nebula-docs/issues/88

Adds Format and Type checking in github actions, similar to https://github.com/DefinedNet/dn-docs/blob/main/.github/workflows/ci.yaml

Authored in gitpod.io, which is why this is coming as a fork rather than direct branch xD

We have "require approval for first time contributors" turned on for running workflows.

image

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -31,3 +31,15 @@ jobs:
         run: npm install
       - name: Check formatting
         run: npm run format:check
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Install node and npm
+        uses: volta-cli/action@v4
+      - name: Install dependencies
+        run: npm install
+      - name: Run tests
+        run: npm run test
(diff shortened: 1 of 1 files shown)


jasikpark:
Not sure why the actions aren't running?

jasikpark:
Oh, needed to enable "run all actions" rather than just DN authored actions.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,45 @@
...
+      - main
+jobs:
+  lint:
+    name: Typecheck
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Install node and npm
+        uses: volta-cli/action@v4
```
IanVS, line 17: We can add `cache: npm` here (and in the other jobs) to speed things up slightly.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,45 @@
...
+      - name: Install dependencies
+        run: npm install
```
IanVS, line 19: Let's use `npm ci`

IanVS (changes requested):
I changed the setting to all outside collaborators for now.  If that becomes a burden we can change it back.

PR #92, opened 2023-01-19
Add some basic descriptions for docs pages

jasikpark:
trying to improve on these listed descriptions mildly

image

### DIFF
--- a/docs/config/logging.mdx
+++ b/docs/config/logging.mdx
@@ -1,6 +1,6 @@
 ---
 sidebar_position: 10
-description: Logging config section
+description: Logging configuration reference for Nebula mesh networking.
 ---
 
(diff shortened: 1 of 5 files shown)


docs/config/logging.mdx
```diff
@@ -1,5 +1,6 @@
 ---
 sidebar_position: 10
+description: Logging config section
```
IanVS, line 3: Should these be something more standalone, like `Logging configuration reference for Nebula mesh networking.`?

jasikpark: That's a better format, thanks 👍

jasikpark: updated w/ that example 🙏

IanVS (approved):

PR #95, opened 2023-03-11
Fix issue number

benyanke:
the lighthouse self-resolve issue has been closed as a duplicate - pointing the page at the current issue to track.


jasikpark (approved):
Thanks for the PR! Nice catch 😄

PR #102, opened 2023-06-26
Update config docs in-line with example config

apalrd:
There are still too many undocumented options in the code, but this at least brings the docs to match the example config and descriptions listed there.

PR #105, opened 2023-07-18
Followup nitpick changes

JohnMaguire:
@apalrd submitted an excellent PR adding documentation around many previously undocumented options in #102. This PR makes a few nitpick changes to things like the meta descriptions, grammar, and adds a couple clarifying notes.

### DIFF
--- a/docs/config/firewall.mdx
+++ b/docs/config/firewall.mdx
@@ -1,6 +1,6 @@
 ---
 sidebar_position: 13
-description: Configure firewall rules for this host
+description: Configuring inbound and outbound firewall rules for a Nebula host
 ---
 
 import { Pill } from '@components/Pill/Pill';
--- a/docs/config/firewall.mdx
+++ b/docs/config/firewall.mdx
@@ -59,9 +59,9 @@ The possible fields of a firewall rule are:
 - `groups`: Same as `group` but accepts a list of values. Multiple values are AND'd together and a certificate must
   contain all groups to pass.
 
-- `cidr`: a CIDR, `0.0.0.0/0` is any.
+- `cidr`: a CIDR, `0.0.0.0/0` is any. This restricts which Nebula IP addresses the rule allows.
 
-- `local_cidr`: a local CIDR, `0.0.0.0/0` is any. This could be used to filter destinations when using unsafe_routes.
+- `local_cidr`: a local CIDR, `0.0.0.0/0` is any. This restricts which destination IP addresses, when using unsafe_routes, the rule allows. If unset, the rule will allow access to the specified ports on both the node itself as well as any IP addresses it routes to. It is a good idea to specify this setting on any box that is configured as an unsafe_routes router. (e.g. Use the node's Nebula IP with a `/32` mask to allow inbound access only to the node itself.)
 
 Logical evaluation is roughly: port AND proto AND (ca_sha OR ca_name) AND (host OR group OR groups OR cidr).
 
--- a/docs/config/firewall.mdx
+++ b/docs/config/firewall.mdx
@@ -1,6 +1,6 @@
 ---
 sidebar_position: 13
-description: Configuring inbound and outbound firewall rules for a Nebula host
+description: Configure inbound and outbound firewall rules for a Nebula host
 ---
 
 import { Pill } from '@components/Pill/Pill';
--- a/docs/config/firewall.mdx
+++ b/docs/config/firewall.mdx
@@ -61,7 +61,11 @@ The possible fields of a firewall rule are:
 
 - `cidr`: a CIDR, `0.0.0.0/0` is any. This restricts which Nebula IP addresses the rule allows.
 
-- `local_cidr`: a local CIDR, `0.0.0.0/0` is any. This restricts which destination IP addresses, when using unsafe_routes, the rule allows. If unset, the rule will allow access to the specified ports on both the node itself as well as any IP addresses it routes to. It is a good idea to specify this setting on any box that is configured as an unsafe_routes router. (e.g. Use the node's Nebula IP with a `/32` mask to allow inbound access only to the node itself.)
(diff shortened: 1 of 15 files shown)


docs/config/static-map.mdx
```diff
@@ -22,13 +22,23 @@ static_map:
...
 <Pill className="mb-24">Default: ip4</Pill>
 
-Select the IP version used to communicate with hosts in the static map. Valid values are `ip`, `ip4`, and `ip6`.
+Select the IP version used to communicate with hosts in the static map. Valid values are  `ip4`, `ip6`, and `ip` (for both.)
+
+:::caution
+
+In general, this should be left as the default value `ip4` to avoid issues communicating with IPv4-only hosts.
+
+Lighthouses learn each node's IP addresses by looking at the source address of incoming packets as well as the self-reported addresses sent by the node. Because most devices are behind a NAT of some sort (e.g. a network router) they cannot self-report their public IPv4 address. By contacting the Lighthouse via its IPv4 address, the Lighthouse is able to learn the node's public IPv4 address.
```
jasikpark: "ipv6 is usually not behind a NAT, so ipv4 is generally the only ip type the lighthouse needs to discover that the host can't self-report"

docs/config/pki.mdx
```diff
@@ -1,6 +1,6 @@
 ---
 sidebar_position: 1
-description: Define the path of each required certificate file
+description: Configuring the node's certificate, private key, CA trust bundle, and blocklist.
```
jasikpark, line 3: Configure the node's certificate

docs/config/preferred-ranges.mdx
```diff
@@ -1,6 +1,6 @@
 ---
 sidebar_position: 8
-description: Set the priority order of underlay IP addresses
+description: Configure the priority order of underlay IP addresses to increase performance.
```
jasikpark, line 3: to increase performance and reduce latency

docs/config/sshd.mdx
```diff
@@ -1,5 +1,6 @@
 ---
 sidebar_position: 11
+description: Enable Nebula's built-in debugging console, accessible via SSH.
```
jasikpark, line 3: accessible via SSH over the Nebula network.

docs/config/firewall.mdx
```diff
@@ -1,6 +1,6 @@
 ---
 sidebar_position: 13
-description: Configure firewall rules for this host
+description: Configuring inbound and outbound firewall rules for a Nebula host
```
jasikpark, line 3:
```suggestion
description: Configure inbound and outbound firewall rules for a Nebula host
```

docs/config/handshake.mdx
```diff
@@ -1,6 +1,6 @@
 ---
 sidebar_position: 16
-description: Handshake Manager Settings
+description: Nebula handshake manager subsystem reference.
```
jasikpark, line 3: what does that blurb mean? "subsystem reference"?

docs/config/pki.mdx
jasikpark, line 3:
```suggestion
description: Configure the node's certificate, private key, CA trust bundle, and blocklist.
```

docs/config/preferred-ranges.mdx
jasikpark, line 3:
```suggestion
description: Configure the priority order of underlay IP addresses to increase performance and reduce latency.
```

docs/config/sshd.mdx
jasikpark, line 3:
```suggestion
description: Enable Nebula's built-in debugging console, accessible via SSH over the Nebula network.
```

docs/config/static-map.mdx
JohnMaguire: I'm unsure about including this level of detail.  I'm trying to keep this a bit more concise because it's an aside.  I've mentioned it a couple other places when chatting with others and it didn't seem necessary to their understanding. How strongly do you feel about this?

jasikpark: fair 👍

jasikpark: not strong

docs/config/firewall.mdx
JohnMaguire: I'm fine with this, but we have a mix of present participle (configuring) and imperative mood (configure).

I think if this was a guide, I would lean harder on "Configure." This is under our reference, so I wasn't really sure what made the most sense.

Do you think we should use imperative mood everywhere?

docs/config/handshake.mdx
JohnMaguire: This is a reference for the handshake manager subsystem? :P

JohnMaguire: It's an awkward one to document because I expect most people will ask the same question: "What are handshake manager settings?" So I aimed to at least clarify that it's a Nebula subsystem.

docs/config/sshd.mdx
JohnMaguire: it doesn't need to be over the nebula network.

docs/config/firewall.mdx
JohnMaguire: Elsewhere we also use "Configuration of" and "X configuration reference"

docs/config/sshd.mdx
jasikpark: oh!

docs/config/handshake.mdx
jasikpark: “Documentation for configuring the nebula handshake manager subsystem” would be clearer to me in that case

docs/config/firewall.mdx
jasikpark: Fair, consistency is nice; “configuring” in particular just stood out as strange to me.

IanVS (approved):

PR #109, opened 2023-10-31
Use corepack pnpm instead of volta npm

jasikpark:
This updates our node/package manager situation to depend on pnpm and corepack, similar to webclient/www


IanVS:
Looks like we need to add `@types/node` to our package.json?

package.json
```diff
@@ -48,11 +49,8 @@
...
     ]
   },
+  "packageManager": "pnpm@8.10.0+sha256.3c5d70d07b0c4849d7e07398b62bf48ca6619ca86f77981125eaab7f5ee82c4c",
   "engines": {
-    "node": ">=16.14"
-  },
-  "volta": {
-    "node": "18.16.0",
-    "npm": "8.17.0"
+    "node": ">=20.9.0"
```
IanVS, line 54: Let's leave this a bit more loose.  No reason not to allow node 18+, is there?

jasikpark: done

jasikpark:
```

  

> nebula-docs@0.0.1 typecheck /home/runner/work/nebula-docs/nebula-docs
> tsc

error TS2688: Cannot find type definition file for '@docusaurus/theme-classic'.
  The file is in the program because:
    Entry point of type library '@docusaurus/theme-classic' specified in compilerOptions
 ELIFECYCLE  Command failed with exit code 2.
```

I can't replicate this locally 🫠

IanVS:
I can.  Maybe try clearing your typescript cache?

IanVS:
This looks like a much bigger PR than just moving from volta to pnpm.  Let's tone it down a bit, and maybe just use regular npm if pnpm is giving us problems.  The main goal is to unblock John.

jasikpark:
> This looks like a much bigger PR than just moving from volta to pnpm. Let's tone it down a bit, and maybe just use regular npm if pnpm is giving us problems. The main goal is to unblock John.

only issue is that corepack doesn't support npm, so looks like John's commit on his own PR is the right way to go then

IanVS:
Alright, then let's get his PR in, then scope this down to just migrating to pnpm?

jasikpark:
reset to "just pnpm"

IanVS (approved):
I pushed a commit to fix the tsc errors.

package.json
```diff
@@ -18,15 +18,18 @@
     "@docusaurus/preset-classic": "2.3.0",
     "@mdx-js/react": "^1.6.22",
     "clsx": "^1.2.1",
-    "parse-domain": "^7.0.1",
```
jasikpark, line 20: why downgrade parse-domain?

jasikpark:
rebased

package.json
IanVS: The last version that wasn't esm-only.

.prettierignore
```diff
@@ -0,0 +1,34 @@
...
+*.seed
+*.pid.lock
+
+# environment variables
+.env
+.env.production
+
+# macOS-specific files
+.DS_Store
+tsconfig.tsbuildinfo
```
IanVS, line 34: Why did we need to add a prettierignore when moving from npm to pnpm?

README.md
```diff
@@ -6,7 +6,25 @@ This is the documentation for the [nebula open source project](https://github.co
...
+requirements set in the `package.json` `engines` field, but you can reference the value in `.nvmrc` for the version that
+is used in CI.
+
+You can install node.js directly from https://nodejs.org/en/download/, or you can use a tool like
+[fnm](https://github.com/Schniz/fnm) to handle automatically changing your node version when switching between projects
+with different requirements.
+
+#### Project dependencies
+
+To install our javascript dependencies, run `pnpm install` (or `pnpm i` for short). You'll want to do this whenever you
```
IanVS, line 28: We'll need to also explain how to get pnpm.

package.json
IanVS: Hm, I think this got lost in a rebase.

.prettierignore
jasikpark: b/c it started formatting `pnpm-lock.json`

README.md
```diff
@@ -6,7 +6,25 @@ This is the documentation for the [nebula open source project](https://github.co
...
+change branches, if there's a possibility that dependencies have been changed or updated. If you're not able to start up
+the client app, this is a good first troubleshooting step to take. To get `pnpm`, run `corepack enable` after installing `node`, and `corepack` should install the right `pnpm` version for you.
```
IanVS, line 30: I'd recommend moving this up higher, for those who are following along step-by-step.

IanVS (approved):
Readme suggestion, but otherwise looks good.

README.md
jasikpark: done

PR #110, opened 2023-10-31
Upgrade to Docusaurus v3

jasikpark:
https://docusaurus.io/blog/releases/3.0


package.json
```diff
@@ -54,6 +54,6 @@
   },
   "packageManager": "pnpm@8.10.0+sha256.3c5d70d07b0c4849d7e07398b62bf48ca6619ca86f77981125eaab7f5ee82c4c",
   "engines": {
-    "node": ">=18.16.0"
+    "node": ">=18.18.2"
```
IanVS, line 57: Just curious, why is this needed?

jasikpark: it's not, i just set a different one accidentally earlier

IanVS: So if `18.16.0` works, we should probably use that, right?  Or honestly just `>=16` if that works.

The downside to using higher versions is that more people are likely to get error messages if they try to contribute, with no real upside so long as the older versions work to build the site.

IanVS (approved):

jasikpark:
i'm a bit confused about why i get a type error locally:

```
❯ pnpm typecheck         

> nebula-docs@0.0.1 typecheck /Users/calebjasik/Git/defined.net/nebula-docs
> tsc

node_modules/@docusaurus/plugin-content-docs/src/sidebars/types.ts:228:3 - error TS2344: Type '"id" | "title" | "frontMatter" | "source" | "sourceDirName" | "sidebarPosition" | "unversionedId"' does not satisfy the constraint '"description" | "id" | "title" | "frontMatter" | "source" | "sourceDirName" | "sidebarPosition" | keyof LastUpdateData | "version" | "slug" | "permalink" | "draft" | "unlisted" | "editUrl" | "tags"'.
  Type '"unversionedId"' is not assignable to type '"description" | "id" | "title" | "frontMatter" | "source" | "sourceDirName" | "sidebarPosition" | keyof LastUpdateData | "version" | "slug" | "permalink" | "draft" | "unlisted" | "editUrl" | "tags"'.

228   | 'id'
      ~~~~~~
229   | 'unversionedId'
    ~~~~~~~~~~~~~~~~~~~
... 
233   | 'sourceDirName'
    ~~~~~~~~~~~~~~~~~~~
234   | 'sidebarPosition'
    ~~~~~~~~~~~~~~~~~~~~~


Found 1 error in node_modules/@docusaurus/plugin-content-docs/src/sidebars/types.ts:228

 ELIFECYCLE  Command failed with exit code 2.
```

but not in CI? not worried about it though, since it's visually fine. just something to sort if it's a problem.

PR #111, opened 2023-10-31
Link to rotation guide from quick start guide

johnmaguire:
This way they have docs to link in their calendar event.


IanVS (approved):
Good thinking.

PR #112, opened 2023-11-16
Fix stats pills styling

IanVS:
|Before|After|
|---|---|
|image|image|

### DIFF
--- a/src/components/Pill/Pill.module.css
+++ b/src/components/Pill/Pill.module.css
@@ -8,6 +8,7 @@
   background-color: var(--ifm-color-emphasis-200);
   padding: 4px 8px;
 
+  &:global(.no-transform),
   & :global(.no-transform) {
     text-transform: none;
   }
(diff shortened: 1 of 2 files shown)


src/components/Pill/Pill.module.css
```diff
@@ -8,6 +8,7 @@
   background-color: var(--ifm-color-emphasis-200);
   padding: 4px 8px;
 
+  &:global(.no-transform),
```
jasikpark, line 11: this seems duplicate?

jasikpark (approved):

src/components/Pill/Pill.module.css
IanVS: I wanted to allow still setting `no-transform` on something inside the pill, as well as on the pill itself.  Though, given this styling problem, maybe that isn't the best idea.  Hopefully MDX/prettier will improve enough to allow us to write it on one line and avoid the extra `<p>` tag some day, though.

jasikpark: oh.. i was forgetting how nesting works, disregard

PR #116, opened 2024-03-19
Added nftables example 

ieugen:
Added nftables example.
I think someone should review the code to make sure it's ok.


johnmaguire:
Superseded by #117

PR #117, opened 2024-03-29
Add ufw and nftables unsafe_routes instructions

johnmaguire:
Thanks @ieugen for the nftables instructions.

### DIFF
--- a/docs/guides/unsafe_routes/index.mdx
+++ b/docs/guides/unsafe_routes/index.mdx
@@ -209,7 +209,52 @@ net.ipv4.ip_forward = 1
 Note: This change is only persistent until you reboot. To make it permanent, add a new line with
 `net.ipv4.ip_forward = 1` to the end of the `/etc/sysctl.conf` file.
 
-### Step 5. Configure iptables on Linux host (LAN)
+### Step 5. Configure packet routing on Linux host (LAN)
+
+Packet routing on Linux can be done using iptables or the newer nftables.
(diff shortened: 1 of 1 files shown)


ieugen:
Nice to see this making it into the docs. 
I would mention that the linux distros are moving to nftables . 
Even UFW is based on nftables after 21.xx 
https://askubuntu.com/questions/1370901/ubuntu-21-10-switched-to-nftables-so-why-is-iptables-still-available 

So having both ufw and nftables config on same system does not make sense IMO. 
I would prioritize the nftables version over the other for the reasons above: everyone is using nftables under the hood. 
Thanks for improving on the config. 
I am quite new to nftables :) .

johnmaguire:
@ieugen Thanks for the feedback. I think we prefer to take an agnostic approach to which firewall users use. I agree it doesn't make sense to configure both ufw and nftables rules. However, if you're using a system like Ubuntu, ufw is still the preferred firewall of choice, to my underrstanding. Mostly, this comes down to which firewall your distro of choice recommends. (We're still missing firewalld instructions. :P)

IanVS (approved):

ieugen:
@johnmaguire : 
I do plan to take a look at firewalld at some point in the future. 
It seems it has received the feature to allow forwarding - but I did not test that https://firewalld.org/2020/04/intra-zone-forwarding . 

Firewalld will generate nftables configuration .

PR #118, opened 2024-04-04
Add aside reccomending Console.app on macOS for logs viewing

jasikpark:
Console.app is a pretty useful tool for watching logs, seems useful to mention.

### DIFF
--- a/docs/guides/viewing-nebula-logs/index.mdx
+++ b/docs/guides/viewing-nebula-logs/index.mdx
@@ -57,6 +57,8 @@ On Intel processors, Homebrew uses `/usr/local` instead of `/opt/homebrew` as a 
 sudo tail -f /opt/homebrew/var/log/nebula.*
 ```
 
+Another useful tool for viewing the logs can be `Console.app` it provides a nice UI for viewing logs in.
+
 ### Service status
 
 ```bash


johnmaguire:
@jasikpark The main purpose of this guide, IMO, is to have a reference point to link people when we want them to collect and send us logs. I'm not sure this information helps us with that story.

If you want to include Console in here, it should probably look more like the Windows Event Viewer story. But the problem with these GUIs that require configuration is that it's a lot more clicks than "Open Terminal and copy and paste this line," and sometimes users get it wrong - like one in the Nebula OSS Slack the other day that missed a checkbox and struggled to get us logs.

So, mostly, I prefer keeping this as simple as possible. I'm open to other viewpoints though.

jasikpark:
That's fair, I'm mostly just thinking of it as helpful for any user debugging their own logs

johnmaguire:
Sorry @jasikpark, I misread which repo this was in. The suggestion seems fine, but maybe we should explain how to find the nebula logs using it?

jasikpark:
How about this?

docs/guides/viewing-nebula-logs/index.mdx
```diff
@@ -57,7 +57,9 @@ On Intel processors, Homebrew uses `/usr/local` instead of `/opt/homebrew` as a
 sudo tail -f /opt/homebrew/var/log/nebula.*
 ```
 
-Another useful tool for viewing the logs can be `Console.app` it provides a nice UI for viewing logs in.
+Another useful tool for viewing the logs can be [`Console.app`](https://support.apple.com/guide/console/welcome/mac) it
```
johnmaguire, line 60: OK, last nitpicks, but I don't think "Another useful tool for viewing the logs can be Console.app it provides a nice UI for viewing logs in" is grammatically correct. Maybe one of the following?

> Console.app can also be a useful tool for viewing logs. To access the logs, ...

> Another useful tool for viewing logs is the built-in Console.app. To access the logs, ....

Further down, I think I usually see shortcuts as 'Cmd + Shift + G' (with or without spaces) instead of 'CMD+SHIFT+G', but this is purely stylistic. Hard to find anything about this, I don't think AP or MLA style covers it. I did find this: https://writing.stackexchange.com/questions/19455/format-keyboard-keys-in-documents

johnmaguire (approved):
Approving as my final comment is a nitpick.

PR #119, opened 2024-04-04
Guide: Debugging with Nebula SSH commands

### LINKED ISSUE #115: Document how to use the built in SSH debug feature
Nebula provides a custom ssh server for debugging connectivity, we should document what this can do and afford

Minimal docs available at https://nebula.defined.net/docs/config/sshd/ already

jasikpark:
Deploy Preview: https://deploy-preview-119--nebula-docs-dn.netlify.app/docs/guides/debug-ssh-commands/

Fixes #115 

This guide attempts to describe how to setup and use the sshd debugability features of nebula. 

I considered adding a section describing each command, but I think it'll be more useful to simply update the help messages for each command to be more useful.

Ref: https://github.com/slackhq/nebula/pull/1109

### DIFF
--- a/docs/guides/debug-ssh-commands/index.mdx
+++ b/docs/guides/debug-ssh-commands/index.mdx
@@ -37,7 +37,7 @@ already have an SSH key for the host you want to access from, follow this guide 
 Set the `listen` property to localhost for local debugging or to your nebula IP to enable access over the nebula overlay
 network.
 
-Next, either send a kill SIGHUP to the nebula PID or restart nebula and ssh via `ssh 127.0.0.1 -p 2222` for example.
+Next, either run `kill SIGHUP <nebula PID>` or restart nebula and ssh via `ssh 127.0.0.1 -p 2222` for example.
 
 Now you can debug your nebula installation, querying things like your hostmap:
 
--- a/docs/guides/debug-ssh-commands/index.mdx
+++ b/docs/guides/debug-ssh-commands/index.mdx
@@ -37,7 +37,7 @@ already have an SSH key for the host you want to access from, follow this guide 
 Set the `listen` property to localhost for local debugging or to your nebula IP to enable access over the nebula overlay
 network.
 
-Next, restart nebula and ssh via `ssh 127.0.0.1 -p 2222` for example.
+Next, either send a kill SIGHUP to the nebula PID or restart nebula and ssh via `ssh 127.0.0.1 -p 2222` for example.
 
 Now you can debug your nebula installation, querying things like your hostmap:
 
--- a/docs/guides/debug-ssh-commands/index.mdx
+++ b/docs/guides/debug-ssh-commands/index.mdx
@@ -81,4 +81,11 @@ list-hostmap - List all known previously connected hosts
     	pretty prints json, assumes -json
 ```
 
+## Notes about some commands
 
+`query-lighthouse <some-ip>` will return an empty result set initially if the host is not connected, but it will trigger
+a background request to the Lighthouse. Meaning, you need to run it twice to actually get a result.
+
+`change-remote` has only a temporary effect: after a period of time, Nebula will "revert" to its preferred remote (i.e.
+based on the rules here:
+https://nebula.defined.net/docs/config/preferred-ranges/#how-nebula-orders-underlay-ip-addresses-it-learns-about)
--- a/docs/guides/debug-ssh-commands/index.mdx
+++ b/docs/guides/debug-ssh-commands/index.mdx
@@ -10,10 +10,11 @@ summary:
 This guide describes useful commands built into the SSH server accessible over nebula, which can allow debugging network
 connectivity for the nebula host.
 
-First you need to configure the [`sshd`](https://nebula.defined.net/docs/config/sshd/) section on the host you want to
-debug.
+First generate a new SSH key for the host you want to debug, via `ssh-keygen -t ed25519 -f ssh_host_ed25519_key`. You
+can set it to only accessible by `root` via `chown root:root ssh_host_ed25519_key`, which will ensure that regular users
+on that host cannot access the private key.
 
-Please follow the instructions here: https://nebula.defined.net/docs/config/sshd/.
+Next configure the [`sshd`](https://nebula.defined.net/docs/config/sshd/) section on the host you want to debug.
 
 Example config:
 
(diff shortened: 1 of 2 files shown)


docs/guides/debug-ssh-commands/index.mdx
```diff
@@ -0,0 +1,82 @@
...
+---
+
+# Debugging with Nebula SSH commands
+
+This guide describes useful commands built into the SSH server accessible over nebula, which can allow debugging network connectivity for the nebula host.
+
+
+First you need to configure the [`sshd`](https://nebula.defined.net/docs/config/sshd/) section on the host you want to debug.
+
+Please follow the instructions here: https://nebula.defined.net/docs/config/sshd/.
```
johnmaguire, line 14: This and the line just above it seem to be stating the same thing. Maybe you can combine them into a single sentence?

"First, update the config of the host you'd like to debug using the instructions found in the [sshd] section of the docs." for example?

The example config below is also copy and pasted from that page. Do you think that's necessary? Maybe it would be better to use this guide as a chance to give more detailed instructions on configuration, rather than copying what exists on the sshd page? e.g. You could walk through how to generate a host key, and an SSH public / private keypair for authentication, in case they don't already have one.

docs/guides/debug-ssh-commands/index.mdx
```diff
@@ -0,0 +1,82 @@
...
+title: Debugging with Nebula SSH commands
+summary:
+  This guide describes useful commands built into the SSH server accessible over nebula, which can allow debugging network connectivity for the nebula host.
+---
+
+# Debugging with Nebula SSH commands
+
+This guide describes useful commands built into the SSH server accessible over nebula, which can allow debugging network connectivity for the nebula host.
+
+
```
johnmaguire, line 11: nit: two line spaces intentional?

docs/guides/debug-ssh-commands/index.mdx
```diff
@@ -0,0 +1,82 @@
...
+list-hostmap - List all known previously connected hosts
+  -by-index
+    	gets all hosts in the hostmap from the index table
+  -json
+    	outputs as json with more information
+  -pretty
+    	pretty prints json, assumes -json
+```
+
+Now you can debug your nebula installation, querying things like your hostmap:
```
johnmaguire, line 75: nit / stye thing, but just because it's "Now you can debug..." I'd consider moving this up before mentioning `help`. Then you can have the `help` bit as the note this guide ends on.

johnmaguire (commented):
This seems like a fine start, but it's largely repeating what already exists on the sshd page. I do think the list of available commands is useful. That being said, it would be even more useful if you could give a little extra color about them, or provide insight into how this tool could be useful.

For example, a note I'm aware of that were shared in recent team meetings: `query-lighthouse <some-ip>` will return an empty result set initially if the host is not connected, but it will trigger a background request to the Lighthouse. Meaning, you need to run it twice to actually get a result.

Or another thing I know if is that `change-remote` has only a temporary effect: after a period of time, Nebula will "revert" to its preferred remote (i.e. based on the rules here: https://nebula.defined.net/docs/config/preferred-ranges/#how-nebula-orders-underlay-ip-addresses-it-learns-about)

docs/guides/debug-ssh-commands/index.mdx
jasikpark: nope, failed to format

docs/guides/debug-ssh-commands/index.mdx
```diff
@@ -0,0 +1,98 @@
...
+    	pretty prints json, assumes -json
+```
+
+## Notes about some commands
+
+`query-lighthouse <some-ip>` will return an empty result set initially if the host is not connected, but it will trigger
+a background request to the Lighthouse. Meaning, you need to run it twice to actually get a result.
+
+`change-remote` has only a temporary effect: after a period of time, Nebula will "revert" to its preferred remote (i.e.
+based on the rules here:
```
johnmaguire, line 97: Maybe instead of the (i.e.) thing just link "preferred remote"?

docs/guides/debug-ssh-commands/index.mdx
```diff
@@ -0,0 +1,98 @@
...
+  authorized_users:
+    - user: steeeeve
+      keys:
+        - '[ssh public key string]'
+```
+
+In this case `steeeeve` and `[ssh public key string]` should be the values for the user you want to enable access for
+debugging. You must add the correct public ssh key for the users you wish to access the ssh server with. If you don't
+already have an SSH key for the host you want to access from, follow this guide by GitHub:
+[Generating a new SSH key and adding it to the ssh-agent](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent).
```
johnmaguire, line 35: nit, but i don't think the ssh-agent part is necessary / relevant. could link https://docs.github.com/en/authentication/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent#generating-a-new-ssh-key

and use "Generating a new SSH key" as text instead.

johnmaguire (approved):
just nits, thanks for adding this!

docs/guides/debug-ssh-commands/index.mdx
jasikpark: Why isn't it relevant? This is for the host that is accessing the nebula node, so you'd want it in your ssh-agent, I agree it's not required for the nebula config side of things.

johnmaguire: It's not necessary to add an SSH key to ssh-agent in order to use it. By default, SSH will use identities stored in `~/.ssh` without adding it to the agent.

Putting it in the agent is mostly useful to avoid repeatedly entering a password, or unlocking a tpm, if that's where you've stored your key.

So it's an optional step, that's not relevant to getting an SSH debug server connected, IMO.

docs/guides/debug-ssh-commands/index.mdx
```diff
@@ -0,0 +1,97 @@
...
+
+Set the `listen` property to localhost for local debugging or to your nebula IP to enable access over the nebula overlay
+network.
+
+Next, restart nebula and ssh via `ssh 127.0.0.1 -p 2222` for example.
```
johnmaguire, line 40: FYI, SSH is reloadable.

PR #123, opened 2024-05-01
Update docs for Nebula v1.9.0

johnmaguire:


docs/config/firewall.mdx
```diff
@@ -75,6 +87,8 @@ firewall:
   outbound_action: drop
   inbound_action: drop
 
+  #default_local_cidr_any: false
```
IanVS, line 90: A user was confused recently, thinking that commented configs represented default values, which is indeed a convention that some projects follow.  Is there any reason this needs to be `false` here, or that it needs to be commented out?

Preference: do not comment this out.  Comments in the docs are almost always (except for Logging, for some reason...) used to indicate actual comments, as opposed to in the example config file where they are used for example configs.

IanVS (approved):
One small suggestion, but these look like good updates to me!

IanVS:
Do we need to wait until 1.9 is released before merging this?

docs/config/firewall.mdx
johnmaguire: I'm a bit confused, because in your first sentence, I think you're referring to confusion in the example config? (In your second paragraph, you note that the behavior here is difference.) If we're thinking of the same user, I responded to that Nebula OSS issue. The user believed that comments in the example config were meant to show defaults - but in actuality, most commented out config values are not defaults, but rather _examples_. (e.g. static_hostmap, and most of the lighthouse section.)

Point taken that the docs are different. I agree that this line could be confusing, but I'm not sure how best to represent it here. Maybe an additional comment above? To answer your question, yes, there's a reason I included it this way, commented out: While the default is `true`, it is only set this way for backwards compatibility. It is most likely in users' best interest to uncomment this line (as it's commented out in the example config) in order to set this to `false` when upgrading to v1.9.0.

The next release will change the default to `false` and deprecate this option. I included this detail in the option's reference section.

What do you think is the best approach here?

IanVS: I think that uncommenting it, perhaps with a comment, is likely the best option for now.  Maybe something like this?

```yaml
default_local_cidr_any: false  # false will become the default in Nebula 1.10.0+
```

PR #125, opened 2024-05-22
Document reloadable configs

IanVS:
Closes https://github.com/DefinedNet/nebula-docs/issues/19

We didn't have any good listing of which options can be reloaded via a SIGHUP vs those that require a full restart to take effect.   This adds a simple badge to indicate which options can be reloaded.  AFAICT, `tun.mtu` is only reloadable on linux, but that's the only difference between platforms that I could find.

### DIFF
--- a/docs/config/pki.mdx
+++ b/docs/config/pki.mdx
@@ -41,6 +41,13 @@ pki:
 
 <Pill className="mb-24">Required</Pill> <Pill className="mb-24">Reloadable</Pill>
 
+:::note
+
+A new certificate will only take effect after a reload if the IP address has not changed, but all other properties of
+the certificate can be changed.
+
+:::
+
 The `cert` is a certificate unique to every host on a Nebula network. The certificate identifies a host’s IP address,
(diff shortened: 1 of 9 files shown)


jasikpark:
Since the default is meant to be "reloadable" wouldn't it make sense to add pills just for the cases where options are "non-reloadable"?

jasikpark (commented):
the design of using a pill sounds fine to me, i'll let others validate that the config options are correctly assigned as reloadable/non-reloadable

docs/config/pki.mdx
```diff
@@ -39,22 +39,24 @@ pki:
 
 ## pki.cert
 
-<Pill className="mb-24">Required</Pill>
+<Pill className="mb-24">Required</Pill> <Pill className="mb-24">Reloadable</Pill>
```
nbrownus, line 42: We should note that while a certificate is reloadable, it will only be accepted if the ip address hasnt changed. All other properties of the cert can change though

nbrownus (approved):
Happy to move forward as it is

PR #128, opened 2024-06-06
Set `onBrokenAnchors` and `onBrokenMarkdownLinks` to `throw`

jasikpark:
- https://docusaurus.io/blog/releases/3.4
	Can enforce that tags are either in the frontmatter or inline w/in content
- https://docusaurus.io/blog/releases/3.3
	Prepares for React 19
- https://docusaurus.io/blog/releases/3.2
	Faster builds
	Faster dev server
	MDX partials TOC (if you add headings via an MDX partial, they can be included in the table of contents now)
- https://docusaurus.io/blog/releases/3.1
	Broken anchors checker

### DIFF
--- a/docusaurus.config.cjs
+++ b/docusaurus.config.cjs
@@ -150,6 +150,7 @@ const config = {
         darkTheme: darkCodeTheme,
       },
     },
+  onBrokenAnchors: 'throw',
 };
 
 module.exports = config;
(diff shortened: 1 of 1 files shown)


IanVS:
Any notable changes in these versions?

jasikpark:
added

IanVS:
It's suggested to set `[onBrokenAnchors](https://docusaurus.io/docs/api/docusaurus-config#onBrokenAnchors)` to `"throw"`, can we try that here?

jasikpark:
done!

PR #129, opened 2024-06-06
Setup dependabot with weekly pnpm and GA updates

jasikpark:
With weekly runs, this should keep nebula-docs up-to-date with minimal interaction and automatic PR creation.

Modelling the slackhq/nebula file: https://github.com/slackhq/nebula/blob/master/.github/dependabot.yml

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,18 @@
+version: 2
+updates:
+  - package-ecosystem: 'npm'
+    schedule:
+      interval: 'weekly'
+    directories:
+      - '/'
+    groups:
+      docusaurus-dependencies:
+        patterns:
+          - '@docusaurus/*'
+      react-dependencies:
+        patterns:
+          - 'react-*'
+  - package-ecosystem: 'github-actions'
+    directory: '/'
+    schedule:
+      interval: 'weekly'


.github/dependabot.yml
```diff
@@ -0,0 +1,18 @@
+version: 2
+updates:
+  - package-ecosystem: 'npm'
+    schedule:
+      interval: 'weekly'
+    directories:
+      - '/'
+    groups:
+      docusaurus-dependencies:
```
jasikpark, line 9: these should group our dependabot and react deps into their own prs

.github/dependabot.yml
```diff
@@ -0,0 +1,18 @@
+version: 2
+updates:
+  - package-ecosystem: 'npm'
```
IanVS, line 3: Should this be `pnpm`?  https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#pnpm

.github/dependabot.yml
```diff
@@ -0,0 +1,18 @@
...
+    schedule:
+      interval: 'weekly'
```
IanVS, line 5: I'd prefer monthly, but if you want to try this for a bit, that's fine.

IanVS (approved):

.github/dependabot.yml
jasikpark: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem

clarifies that it's supported via the `npm` yaml value

PR #141, opened 2024-06-07
Color-code Nebula and regular IPs

jasikpark:
- **Start color coding overlay and underlay ips**
- **More color coding**
- **Even more color coded overlay/underlay ip addresses**


IanVS:
I'm not 100% sure about using underlines here, it makes them look like links to me.  And the color difference is pretty subtle / easy to miss.  I'd want to have @bgolat take a look and give his take, at least.  Brian, a good page to look at might be https://deploy-preview-141--nebula-docs-dn.netlify.app/docs/config/lighthouse/.

jasikpark:
The styles are 100% up for debate, I just wanted to get something semi-reasonable up

bgolat:
What is the context of this?  I would avoid underlining text as a way to highlight since that it generally reserved for links.

jasikpark:
Highlighting would probably be preferred then; the goal is to indicate when an IP or address refers to a Nebula overlay IP or an underlay IP from the regular network.

PR #223, opened 2025-03-17
Add guide to upgrading to cert-v2

jasikpark:
Hopefully this is a good starter for folks trying to upgrade their network.

Deploy Preview: https://deploy-preview-223--nebula-docs-dn.netlify.app/docs/guides/upgrade-to-cert-v2-and-ipv6/

- [ ] Validate loading a v1cert from a v2 CA on a pre-v2cert nebula install, possibly recommend that v1 certs are signed by v1 ca's and v2 certs are signed by v2 ca's.

### DIFF
--- a/docs/guides/upgrade-to-cert-v2-and-ipv6/index.mdx
+++ b/docs/guides/upgrade-to-cert-v2-and-ipv6/index.mdx
@@ -139,15 +139,12 @@ on v2 only. For a small network, it may be fine to simply change everything over
 After each step, you may validate that your changes have taken effect by using the `print-tunnel` command within the
 [debug ssh server](/docs/guides/debug-ssh-commands/).
 
-For example:
+For example, when debugging `nebula`'s tunnel to `lighthouse1` where both have both v1 and v2 certs and
+`pki.default_version` is set to `1` on both hosts, we can see that the certificate used in the tunnel is version 1.
 
 ```bash
-calebjasik@nebula > print-tunnel 192.168.1.1
-{"vpnAddrs":["192.168.1.1"],"localIndex":2965094606,"remoteIndex":1285130098,"remoteAddrs":["172.17.0.3:4242"],"cert":{"details":{"curve":"CURVE25519","groups":[],"isCa":fa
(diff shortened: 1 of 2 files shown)


jasikpark:
Now that https://github.com/slackhq/nebula/pull/1359 is merged, the docs can be updated to reference the added logs

docs/guides/upgrade-to-cert-v2-and-ipv6/index.mdx
```diff
@@ -0,0 +1,279 @@
...
+
+The basic steps go:
+
+import TOCInline from '@theme/TOCInline';
+
+<TOCInline toc={toc} />
+
+## Upgrade to nebula nightly
+
+First, download the updated nebula version that has v2 certs and IPv6 support from nightly releases, and move all the
```
johnmaguire, line 20:
```suggestion
First, upgrade each host in your network to a nebula version with support for v2 certs. Currently this is only available in the nightly releases.
```

Maybe add a note on how to check? i.e. `nebula -version`? What version string should they see?

johnmaguire (commented):
Thanks for putting this together! I haven't quite finished going through it, but I wanted to drop some early feedback. If you take the suggestions, maybe read on through the rest of the guide yourself and see if we can clean anything else up as a result.

When writing technical docs like this, I aim to be precise yet concise, so in some cases I'm advocating for removing duplicate information.

docs/guides/upgrade-to-cert-v2-and-ipv6/index.mdx
```diff
@@ -1,147 +1,161 @@
...
 }
 ```
 
-Copy the public key of the CA and add it to the `pki.ca` list in the config for every nebula host:
-[https://nebula.defined.net/docs/config/pki/#pkica](https://nebula.defined.net/docs/config/pki/#pkica)
+Append the new CA certificate to the [`pki.ca` trust bundle](https://nebula.defined.net/docs/config/pki/#pkica) in the
+config of every Nebula host and reload Nebula.
 
-## Re-issue v1+v2 certificates for every host with the v2 CA
+## Issue v1+v2 certificates bundles
```
jasikpark, line 68: maybe this should be "certificate" not "certificates"?

PR #261, opened 2025-08-06
Fix links in footer

IanVS:


johnmaguire (approved):

docusaurus.config.cjs
```diff
@@ -123,10 +123,6 @@ const config = {
                 label: 'Docs',
                 href: 'https://docs.defined.net',
               },
-              {
-                label: 'Twitter',
```
jasikpark, line 127: wanna replace this with bluesky?

PR #301, opened 2025-12-09
Document pki.initiating_version configuration

jasikpark:
Add documentation for pki.initiating_version and its default behavior.

Fixes https://github.com/DefinedNet/nebula-docs/issues/222

Is there other new config in v1.10? IDK.

### DIFF
--- a/docs/config/pki.mdx
+++ b/docs/config/pki.mdx
@@ -86,7 +86,9 @@ stolen or compromised.
 
 <Pill className="mb-24">Default: 0</Pill>
 
-`pki.initiating_version` determines which nebula cert version to use when initiating new connections. If unset (or set
-to 0), it will default to v1 if any v1 cert is available, otherwise it will default to v2.
+`pki.initiating_version` determines which nebula cert version to use when initiating new connections. This setting only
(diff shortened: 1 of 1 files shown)


docs/config/pki.mdx
```diff
@@ -81,3 +81,12 @@ stolen or compromised.
 <Pill className="mb-24">Default: False</Pill> <Pill className="mb-24">Reloadable</Pill>
 
 `disconnect_invalid` is a toggle to force a client to be disconnected if the certificate is expired or invalid.
+
+## pki.initiating_version
```
johnmaguire: This setting is not applicable most of the time, and only applies if both cert v1 & v2 are defined (i.e. in a transitional state.) Here's the docs from the Nebula config which clarify this and how to use it in practical terms:

```
  # This setting only applies if both a v1 and a v2 certificate are configured, in which case it will default to `1`.
  # Once all hosts in the mesh are configured with both a v1 and v2 certificate then this should be changed to `2`.
  # After all hosts in the mesh are using a v2 certificate then v1 certificates are no longer needed.
```

Maybe just use that?

PR #303, opened 2025-12-13
Starlight

jasikpark:
Exploring moving to https://starlight.astro.build/

See: https://starlight.docs-nebula.pages.dev/

Mostly pretty smooth, I haven't ripped out everything docusaurus yet.

- The asides have the same color on light and dark mode, and don't blend well, it might be smart to just use their default colors
- We can replace the header entirely, if we wish. I think it's fine leaving as-is on nebula-docs.
- We will probably want to add a custom footer, to replicate the existing links in the footer


jasikpark:
Closing — abandoned Starlight migration draft, untouched since December, now conflicting.

PR #350, opened 2026-03-23
Update Nebula docs with new settings

johnmaguire:


jasikpark (approved):
lgtm, from a quick glance

PR #360, opened 2026-06-16
Update ICMP message for rejected packets

JackDoan:


jasikpark:
Verified this against the merged Nebula change ([slackhq/nebula#1766](https://github.com/slackhq/nebula/pull/1766)) — the IPv4 reject packet now sends ICMP type 3 / code 13, so the correction is accurate. 👍

One small consistency nit: Nebula's own source comment and `examples/config.yml` both phrase code 13 as "Communication administratively prohibited". This wording drops "Communication". Suggest matching the upstream phrasing exactly so the two sources line up:

> For other protocols, this will be an ICMP "Destination unreachable: Communication administratively prohibited" packet.

(IANA's official name for type 3 / code 13 is likewise "Communication Administratively Prohibited".)

jasikpark (approved):

PR #362, opened 2026-08-03
Add host discovery guide (static_host_map vs advertise_addrs)

jasikpark:
**Deploy preview:** https://push-mpkswpmozxnl.docs-nebula.pages.dev/ — new guide at [/docs/guides/host-discovery/](https://push-mpkswpmozxnl.docs-nebula.pages.dev/docs/guides/host-discovery/)

Adds a concept guide, How Hosts Find Each Other, explaining how a host learns a peer's routable underlay address: the static bootstrap path (`static_host_map`, dialed directly, works without lighthouses) versus lighthouse-mediated dynamic discovery (self-reported addresses, source-address learning for NAT'd hosts, `advertise_addrs`) — and when to use each, including why lighthouses must appear in every host's static host map while their own stays empty.

Supporting edits:

- `docs/config/static-host-map.mdx` — bootstrap framing, mixed IP/DNS name/IPv6 example, `static_map` cross-link, link to the guide
- `docs/config/lighthouse.mdx` — contrast paragraph in the `advertise_addrs` section (set once on the host vs edit every peer's config)
- `docs/guides/quick-start/index.mdx` — links the empty-lighthouse `static_host_map` example to the guide's explanation

Includes a Managed Nebula callout following the precedent in `intro.md` and the rotating-CA guide. `pnpm build` passes, so all new cross-links resolve.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01AU3rHtCYz5p8iNy5krguXn

### DIFF
--- a/docs/config/lighthouse.mdx
+++ b/docs/config/lighthouse.mdx
@@ -52,8 +52,8 @@ configured to be lighthouses in your network
 <Pill className="mb-24">Default: False</Pill>
 
 `serve_dns` optionally starts a DNS listener that responds to `A` and `TXT` queries and can even be delegated to for
-name resolution by external DNS hosts. To enable listening on IPv6 in addition to IPv4, set `lighthouse.dns.host` to
-`'[::]'`.
+name resolution by external DNS hosts. To enable listening on IPv6 in addition to IPv4, set
+[`lighthouse.dns.host`](#lighthousedns) to `'[::]'`.
 
 The DNS listener can only respond to requests about hosts it's aware of. For this reason, it can only be enabled on
 Lighthouses.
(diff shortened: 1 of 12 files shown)


bgolat (dismissed):
Reviewed the docs changes. The "new network" qualification and the enabling/upgrading guide are accurate — the DNClient v0.9.3 requirement and the "Use managed lighthouses" admin-panel label were both confirmed correct.

Non-blocking polish suggestions for a follow-up:
- quick-setup.mdx: "the managed lighthouse covers this for you out of the box" is still unqualified and slightly contradicts the new IPv4-only guidance.
- managed-lighthouses/index.mdx: the "Enabling the managed lighthouse" section repeats the auto-provisioning point from "What you get automatically," and the heading buries the one audience (existing dual-stack networks) that actually needs to act.

Approving — these are optional.

bgolat (approved):
Approving — nice addition, the host-discovery guide fills a real gap and the cross-linking is a good cleanup.

One nit (non-blocking): in `docs/config/static-host-map.mdx`, the new intro says *"Every host must have an entry for each lighthouse in the network."* That's slightly overstated — lighthouses themselves are the exception, which the new guide itself notes ("a lighthouse's own `static_host_map` is typically empty, and multiple lighthouses generally don't need to know about each other"). Scoping it to non-lighthouse hosts (or "every host that needs to reach a lighthouse") would keep the reference page consistent with the guide. Fine to fix in a follow-up.

PR #376, opened 2026-08-26
Document offload and CPU pinning config

JackDoan:
maybe don't merge until 1.12 is cut

### DIFF
--- a/docs/config/listen.mdx
+++ b/docs/config/listen.mdx
@@ -37,6 +37,10 @@ nodes. Using 0 on lighthouses and relay hosts will likely lead to connectivity i
 
 Sets the max number of packets to pull from the kernel for each syscall (under systems that support `recvmmsg`).
 
+When [`listen.udp_offloads`](#listenudp_offloads) has enabled GRO (Linux kernel 5.10+), each receive slot is sized for a
+full 64KiB coalesced superpacket, so the receive scratch memory is `batch × 64KiB` per listening socket (about 4MiB per
+routine at the default of 64). Lower this value to trade peak per-syscall throughput for memory on constrained hosts.
+
 ## listen.read_buffer, listen.write_buffer
 
 Configure socket buffers for the udp side (outside), leave unset to use the system defaults. Values will be doubled by
@@ -91,3 +95,32 @@ ip rule add not from all fwmark 4242 lookup 4242
 ip rule add from all lookup main suppress_prefixlength 0
 ip route add default dev nebula1 via <nebula_unsafe_route_g> table 4242
 ```
+
+## listen.rebind_on_network_change
+
+<Pill className="mb-24">Default: True</Pill>
+<Pill className="mb-24">Added in v1.11.0</Pill>
+
+This option is only supported on macOS and is not reloadable.
+
+macOS scopes the UDP socket to the interface it was created on, so moving between networks (wifi to wired, office to
+home) leaves Nebula sending out an interface that no longer has a route. When enabled, Nebula watches the routing socket
+and rebinds the listener once the change settles, so devices moving between networks recover without waiting for dead
+tunnel detection. iOS does not use this option, the host app drives the same rebind itself.
+
+## listen.udp_offloads
+
+<Pill className="mb-24">Default: False</Pill>
+<Pill className="mb-24">Added in v1.12.0</Pill>
+
+This option is only supported on Linux and is not reloadable.
+
+When enabled, Nebula attempts to enable UDP Generic Segmentation Offload (GSO, Linux kernel 4.18+) and Generic Receive
+Offload (GRO, Linux kernel 5.10+) on its UDP socket(s). With GSO, batches of packets bound for the same host are handed
+to the kernel as a single "superpacket" along with the size to split it back into individual datagrams, reducing
+per-packet syscall and network stack overhead on the send path. With GRO, the kernel coalesces bursts of datagrams
+arriving from the same host into a single large read. Support for each feature is probed when the socket is created, and
+Nebula falls back to standard sends and receives on kernels that lack it.
+
+Enabling GRO increases the fixed memory used by the receive path, see [`listen.batch`](#listenbatch) for details. For
+the biggest throughput gains, pair this option with [`tun.use_offloads`](/docs/config/tun/#tunuse_offloads).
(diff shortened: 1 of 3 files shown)


jasikpark (commented):
Read through this against the nebula source. The writing is clear and the tricky parts (the batch × 64KiB math, the TSO/USO fallback ladder, the `cpu_affinity` validation and modulo-cycling) all check out. Three things I'd fix before merging.

**1. `tun.pin_threads_key` doesn't actually work as documented.** The default is documented as `pid`, but `main.go` logs the chosen key and then unconditionally overwrites it with the bound UDP port, so `pid` is unreachable on any node with a bound port. That's a code bug rather than a docs bug — I opened slackhq/nebula#1877 to gate the port lookup on the key. If that lands before 1.12 this page is correct as written; if it doesn't, the default needs to say `port`.

**2. "It only takes effect when `routines` is greater than `1`" is wrong for `tun.pin_threads` and `tun.cpu_affinity`.** `listenIn` calls `pinThisThread` for every queue whenever `pinThreads` is true, including the single-routine case (it pins to `allowed[0]`). What's gated on `routines > 1` is only the *automatic* `cpupick.Default` selection. So with `routines: 1` and `pin_threads: true` you do get pinned, and an explicit `cpu_affinity` list still applies. Suggest narrowing that sentence to the automatic-selection paragraph, and softening the same slant in the routines.mdx addition ("When set above `1`, the reader threads are pinned to CPUs by default" reads as though `1` isn't pinned).

**3. `listen.batch: 1` silently disables GRO.** `udp_linux.go` only calls `prepareGRO()` when `s.Batch > 1 && s.Offloads`. The new `listen.batch` paragraph invites people to lower the value to save memory without mentioning that `1` turns the receive-side offload off entirely. Worth a sentence there.

Two smaller notes:

- Worth saying explicitly on the `listen.batch` addition that the GRO memory behavior is Linux-only, since `listen.batch` itself is documented generically for anything with `recvmmsg` and the platform caveat lives on the other page.
- `tun.cpu_affinity` and `tun.pin_threads_key` don't list a platform pill the way the others do — they say Linux-only in prose, which is fine, just checking that's deliberate for consistency.

jasikpark (commented):
Correction to my review above — my nebula checkout was stale, so two of my three points were wrong. Sorry for the noise. I re-verified everything against `slackhq/nebula@abff1a4` (current master).

**Point 1 (`tun.pin_threads_key`) — withdraw, already fixed.** slackhq/nebula#1861 fixed the unconditional port override back in August; my checkout predated it. The `port` lookup is now correctly inside `case "port"`. **The page is accurate as written, `pid` really is the effective default.** I opened a redundant PR for this and have closed it.

**Point 2 (`routines > 1`) — withdraw, I had it backwards.** `NewInterface` coerces before copying the field:
```go
if c.routines <= 1 {
    c.PinThreads = false //pinning is not useful unless there's more than one tun reader
}
```
(`interface.go:201-203`, landed with the feature in #1704.) So at `routines: 1`, `f.pinThreads` is false, `pinThisThread` never runs, and an explicit `cpu_affinity` list is stored but inert. **The PR's "only takes effect when routines is greater than 1" is correct for both options** and matches `examples/config.yml` ("Coerced to false if routines <= 1"). The routines.mdx sentence is fine too.

One genuine edge case if you want to be exhaustive, though I'd probably skip it: the coercion keys off the *configured* routines, while `activate()` separately clamps `f.routines` to 1 when the UDP side lacks multi-reader support. So `routines: 2` on such a platform keeps `pin_threads` true and pins the single remaining queue. Niche enough not to be worth documenting.

**Point 3 (`listen.batch: 1` disables GRO) — still stands, confirmed.** `udp_linux.go:80` is `if s.Batch > 1 && s.Offloads { out.prepareGRO() }`, and that's the only `prepareGRO` call site. Nothing is logged when it's skipped, so `batch: 1` with `udp_offloads: true` silently loses receive-side offload. Since the new paragraph specifically suggests lowering `batch` to save memory, this is worth a sentence. Note the gate is GRO-only — GSO is still enabled at `batch: 1`, since `s.Offloads` is passed to `newBatchWriter` unconditionally.

Two small refinements to the new text while we're here, both on things I'd originally waved through:

- "Support for each feature is probed when the socket is created, and Nebula falls back to standard sends and receives on kernels that lack it." Two nits. The send-side fallback is still `sendmmsg` with one entry per datagram rather than "standard sends", and fallback isn't only at creation: an `EIO` on a superpacket disables GSO at runtime and replays the batch, because the setsockopt probe can't see per-route limitations. Maybe "...falls back to unbatched sends and receives on kernels that lack it, and disables GSO at runtime if the kernel later rejects a superpacket."
- The batch math is right (`65535` per slot plus a 28-byte sockaddr and 24-byte cmsg, so 4.003 MiB at batch 64), and per-socket is genuinely per-routine since each routine gets its own `SO_REUSEPORT` listener. Worth noting the non-GRO path allocates `batch × 9001` (~0.55 MiB) for contrast, which makes the tradeoff concrete.

And the point from my last comment that still stands: `listen.batch` is documented generically for any `recvmmsg` system, so the GRO memory paragraph on that page should say Linux-only explicitly. Also worth knowing these are Linux-*and-not-Android*, and on every other platform both options are silently accepted and ignored with no warning.

Net: the PR is in better shape than my first pass suggested. Only #3 and the wording nits are real.

jasikpark (commented):
Checked this against `slackhq/nebula@abff1a4`. Accurate throughout, including the batch memory math and the platform caveats. One thing worth adding:

**`listen.batch: 1` silently disables GRO.** `udp_linux.go:80` is the only `prepareGRO` call site and reads `if s.Batch > 1 && s.Offloads`, with nothing logged when it's skipped. Since the new `listen.batch` paragraph invites lowering the value to save memory, worth noting that `1` turns receive-side offload off entirely. The gate is GRO-only, GSO still applies at `batch: 1`.

Two optional wording nits on `listen.udp_offloads`, both about "probed when the socket is created, and Nebula falls back to standard sends and receives":

- The send-side fallback is still `sendmmsg` with one entry per datagram, so "unbatched sends" is more precise than "standard sends".
- Fallback isn't only at creation. An `EIO` on a superpacket disables GSO at runtime and replays the batch, since the setsockopt probe can't see per-route limitations.

